Issue #8, opened 2015-10-24
Complete adding descriptions to extension manifests


anaran:

anaran:
I'm already doing this

wbamberg:
Closed by PR #9 . Thanks @anaran .

PR #9, opened 2015-10-24
Fixes issue #8

anaran:


wbamberg:
Thanks for this. We definitely need descriptions (really we should include everything that's not mandatory, but is nevertheless good practice).

Question, though: do you think the descriptions should be from the point of view of the user of the add-on, or from the point of view of the developer? For example, for notify-link-clicks, should we have something like:

(1) "Injects a content script into all pages: the content script sends a message to the background script when the user clicks links. The message contains the link's href. When the background script gets the message, it displays the href using the notifications API."

or

(2) "Displays notifications when the user clicks links."

I think, actually, it should be (2). It's shorter, less likely to break when we update the add-on, and we already document the internals somewhere else. Better, maybe:

(3) "Displays notifications when the user clicks links. See https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Examples#notify-link-clicks."

beastify/manifest.json
```diff
@@ -1,5 +1,6 @@
 {
 
+  "description": "Adds a browser action icon to toolbar to select a beast from popup menu to replace document body content with an image of the beast in active tab.",
```
wbamberg: This sentence is pretty long. Could we say: "Adds a browser action icon to the toolbar. Click the button to choose a beast. The active tab's body content is then replaced with a picture of the chosen beast."

wbamberg:
Thanks @anaran !

PR #11, opened 2015-10-27
Create page action webextensions example from 1197422#c27

### LINKED ISSUE #10: Add a demo for page_action
I have one ready, based on
[1197422 – Implement pageAction API for open extension API](https://bugzil.la/1197422#c27)

Took out the browser_action to make it strictly a page_action demo.

Should be handy to verify a followup bug I submitted,

[1218175 – pageAction extension no longer works after browser restart](https://bugzil.la/1218175#c1)

anaran:
Fixes #10 once :kmag approves.


shaggy-page-action/popup.html
```diff
@@ -0,0 +1,8 @@
+<!DOCTYPE html>
+<html>
+  <head>
+  </head>
```
wbamberg: I think you should declare the charset here or you will get an error in the console.

wbamberg:
Thanks so much for adding a new example! I did have a couple of comments:
* is it too paranoid to worry about the licensing of the image it includes?
* since the point of a page action is that it only appears on some pages, it seems a pity that this always shows the button for the active tab. Could we instead have some condition in the page that we use to decide whether to show it?

Before this PR, I was experimenting with a couple of options:

(1) show the page action when there are bookmarks whose URL has the same host (using the bookmarks API), and the page action shows a popup listing those bookmarks. So if you are on a page in Wikipedia, you can quickly see all the other Wikipedia pages you have bookmarked.

(2) show the page action when the page sets tracking cookies (using the webReqest API, probably). List the tracking cookies when it is clicked.

But both of these seemed to be a bit too involved to make good examples (we don't support bookmarks.search yet, and even if we did, it seems like you can't search with wildcards).

But even just showing the icon when the URL matches a particular pattern might be good enough?

anaran:
Hi Will, I share your views on what you wrote above.

:kmag just confirmed o IRC we probably shouldn't be using that image.

A rename of the example is probably in order too. What should we go for?

tumbeasts are CC licensed. We could follow https://bugzilla.mozilla.org/show_bug.cgi?id=704085 and add a LICENSE file.

---

I guess we should define the scope of these examples, as PR #13 also shows (e.g. build steps).

The scope could be specified in README.md (or perhaps in a TBD CONTRIBUTING.md) as well as at the top of https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Examples with minimal duplication.

For me these examples are not only a good goto location for developers, but also casual users which I see as prospective developers, given a pleasant experience with the examples.

My other favourite direction is to minimize confusion or lack of context for users who might encounter these examples unexpectedly, like finding them installed in about:addons or Gaia Settings->Add-ons.

That's why I tried my hand at issue #8.

Explorability is a concern as well, as we discussed and you addressed nicely with the new-beast in 561eb1d498926accd6b31df3b6c69274060a421d.

wbamberg:
> A rename of the example is probably in order too. What should we go for?

I think it depends on what we want it to do :-). If it's going to show tumbeasts, then "show-tumbeasts" would work.

I did have a couple of ideas for page action examples, but couldn't really get either of them off the ground. Writing them here in case they can inspire anything else.

* a page action that's shown when you have bookmarked any pages on the same host. Click the button to show those bookmarks. So when I'm at wikipedia.org, it will show me any other wikipedia pages I have bookmarked. It's not terribly useful, but not totally useless either, and shows the bookmarks API as well. But (1) bookmarks.search() is not yet supported in Firefox, and (2) on Chrome, it seems like you can't use wildcards in the URLs passed into bookmarks.search().

* a page action that's shown when the page or any resources it loads sets persistent (tracking) cookies. Click the button to show the cookies and delete them. I think the tricky bit here seems to be figuring out which cookies get set by which page. I did manage to use https://developer.chrome.com/extensions/webRequest#event-onHeadersReceived to log persistent headers, and you can get the tabID from that. But it ends up getting a bit involved for a simple example. Still, I might go back to this some time.

A dead-simple example, that just displays the page action based on some easy-to-test conditional, and just shows an image, would be useful too though.

> I guess we should define the scope of these examples, as PR #13 also shows (e.g. build steps).
> 
> The scope could be specified in README.md (or perhaps in a TBD CONTRIBUTING.md) as well as at > the top of https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Examples with minimal
> duplication.

Great idea. I also think CONTRIBUTING.md is a very very good plan. I'll file an issue for it :-).

For me, the primary function of these examples is to give developers examples of use of the APIs. My ultimate goal would be that if you're looking at the docs for, say, "tabs.executeScript", you can click a link to see all the examples for that API, in an executable form. The idea being that executable examples like this are a good complement to examples that are just text in the docs page. (In a proper sci-fi future, this would all be automated: so a script could analyse the examples to figure out which APIs are used in which examples, then a page could query to get the set of examples that demonstrate a particular API.)

I think any other function of the examples is secondary. I like the idea of it appealing to casual users too, and if we can make the examples engaging for them, that's great too. But if that makes the examples include too much nonessential complexity, then I'd be reluctant to do that.

> My other favourite direction is to minimize confusion or lack of context for users who might
> encounter these examples unexpectedly, like finding them installed in about:addons or Gaia 
> Settings->Add-ons.

I'm not sure how a user might encounter them unexpectedly? But I do agree, that descriptions and such are good. I was meaning to comment on your PR for that, and will do it now :-).

wbamberg:
This PR has had no activity in almost a year, and we now have a couple of page action examples, so I'm closing it.

PR #12, opened 2015-10-27
Add new communicate demo (cs to bg with async responses)

anaran:
Hi Will!

I wrote this one myself.

I'm not too fond of the `communicate` name myself, but nothing better and still short sprung to mind.

It demonstrates how the content script can obtain extension information from the background script by sending messages and receiving asynchronous responses.

I would love to find a way not having to inject the content script UI dynamically.

On the content script side there does not seem to be a way.

I'll need this for some jetpack add-ons I plan on porting once webextensions options will become available (see [1218472 – Implement options API for open extension API](https://bugzil.la/1218472)).

### DIFF
--- a/communicate/communicate-bg.js
+++ b/communicate/communicate-bg.js
@@ -6,34 +6,34 @@ function handleMessages(message, sender, sendResponse) {
 	switch (message.type) {
 		case "tabs":
 			chrome.tabs.query({currentWindow: true}, function(tabs) {
-				sendResponse ({ tabs: tabs });
+				sendResponse({ selection: message.selection, tabs: tabs });
 			});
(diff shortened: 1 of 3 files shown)


anaran:
The problem with moving the extension's UI to a static page opening in a new tab (thereby avoiding "accidental complication") is how to summon up that tab from the active tab.

I am trying to avoid the browser action and page action UI in this example since it will most likely not be available on Android and Firefox OS.

See also https://github.com/anaran/anaran-jetpack-content/blob/master/README.md which is about jetpack, but takes the same line.

wbamberg:
Closing, as this PR hasn't seen any activity in a while.

PR #13, opened 2015-11-02
Add the first cut at the ui-playground.

bwinton:
@wbamberg: I suspect there will be a bunch of cleanup needed for this, and I'm not sure how you feel about build scripts in add-ons, but I thought I would post this up, and see how you feel about it.  :grinning:


wbamberg:
@bwinton : I guess my worry about using build scripts is that we're adding another non-essential thing people have to learn in order for them to learn about WebExtensions. The counter to that is (1) these are useful things for people to learn anyway, and maybe even (2) these are just fundamental skills that we expect people to have already, like JavaScript.

For example, we assume it's OK to expect people to have basic Git skills to be able to access the examples. Can we assume the same about npm and Gulp? I'm not sure. Maybe. Either way, I think this needs a README explaining to people how to get started with this.

Also, these kinds of examples have more value (to me) if there is a MDN article talking about the technique they demonstrate (for instance, I'd like us to have i18n support for some of these, but that's most useful when we can have an article talking about i18n, that points to the examples. Then they reinforce each other. So maybe we should have an article on using build scripts, or even more specific techniques like using build scripts to support ES6?

I think gulpfile.js has a problem at line 59, doesn't it [1]?

Anyway, I tried following the instructions from the Gulp getting-started[2] and got:

    Error: Cannot find module 'gulp-babel'

So I installed gulp-babel, and of course then got:

    Error: Cannot find module 'merge-stream'

Am I doing something wrong?

[1] https://github.com/mdn/webextensions-examples/pull/13/files#diff-90eb2726ed6a723d8e69f3afd29fa268R59
[2] https://github.com/gulpjs/gulp/blob/master/docs/getting-started.md#getting-started

bwinton:
Yeah, I sort of agree.  I was using a build script mostly because it makes my life easier (at least until we can [load unpacked add-ons](https://bugzilla.mozilla.org/show_bug.cgi?id=1209344) 😉), but it wasn't really necessary, so I've removed it.  I also added a small readme.  Lemme know if there's any way I could improve it!

wbamberg:
My most general comment about this extension is that it does a lot of disparate things, and it's not obvious what it's doing and why. In general it's nice for these examples to do one clearly defined and easily graspable thing, that way you can see how the APIs can be used to accomplish particular things.

I understand that this isn't that sort of example, and that's fine, but I think it then needs more comments and/or README explaining a bit what it's all doing.

ui-playground/lib/background.js
```diff
@@ -0,0 +1,114 @@
+/* globals chrome:false */
+
+'use strict';
+
+chrome.browserAction.setBadgeText({text: '🚲ᕕ( ᐛ)ᕗ'});
+// chrome.browserAction.setBadgeBackgroundColor({color: '#FFF'});
```
wbamberg: Commented-out code should be removed.

ui-playground/lib/background.js
```diff
@@ -0,0 +1,114 @@
...
+
+var chillOut = function (aTab) {
```
wbamberg: Maybe just me, but I found this function quite hard to read. As I understand it, it's doing this:
* if the active tab has "chilloutand..." loaded already, do nothing
* if some other tab has "chilloutand..." loaded already, switch to it
* otherwise, open a new tab with "chilloutand..." loaded

To me, code like this would be easier to understand:

    var chillOut = function (aTab) {
      console.log("Chilling…");
      var catGifs = 'http://chilloutandwatchsomecatgifs.com/';

      chrome.tabs.query({'url': catGifs}, (tabs) => {
        if (tabs.length) {
          for (var tab in tabs) {
            if (tab.active) {
              return;
            }
          }
          chrome.tabs.update(tabs[0].id, {'active': true});
        }
        else {
          chrome.tabs.create({'url': catGifs, 'active': true}, (tab) => {
            console.log(tab);
          });  
        }
      });
    }

But maybe this is a just a personal thing, or I'm missing something.

ui-playground/lib/background.js
```diff
@@ -0,0 +1,114 @@
...
+        // chrome.pageAction.show(tab.id);
+      });
+    }
+  });
+};
+
+var count = 0;
+
+chrome.notifications.onClicked.addListener(function (notificationId) {
+  // This doesn't work yet.
```
wbamberg: I don't think we should include things that don't work.

ui-playground/lib/background.js
```diff
@@ -0,0 +1,114 @@
...
+  console.log("Notification Clicked!!!", notificationId);
+  chrome.tabs.query({active: true}, function (aTab) {
+    chillOut(aTab);
+  });
+});
+
+var audio = new Audio();
```
wbamberg: I don't think any of this audio stuff works, because we don't support bookmarks.onCreated. I think examples should only contain things that work in their target version of Firefox.

ui-playground/lib/background.js
```diff
@@ -0,0 +1,114 @@
...
+        audio.play();
+      }
+    }
+  });
+}
+
+chrome.alarms.onAlarm.addListener(function (alarm) {
+  console.log("Alarm!!!", alarm, count);
+  chrome.browserAction.setBadgeBackgroundColor({color: '#' + Math.floor(Math.random() * 255 * 255 * 255).toString(16)});
+  count = (count + 1) % (Math.floor(Math.random() * 10) + 1);
```
wbamberg: I found this stuff with "count" curious. To me it's extra stuff someone has to understand, that doesn't help them actually use the APIs. Could we not just do this every time the alarm fires?

ui-playground/lib/background.js
```diff
@@ -0,0 +1,114 @@
...
+    });
+    chrome.tabs.query({active: true}, function (aTab) {
+      chillOut(aTab);
+    });
+  }
+});
+
+chrome.alarms.create('chillOut', {periodInMinutes: 1});
+if (chrome.pageAction) {
+  chrome.pageAction.onClicked.addListener(function (aTab) {
```
wbamberg: This will never happen because pageAction has a popup.

ui-playground/lib/background.js
```diff
@@ -0,0 +1,114 @@
...
+      isClickable: true
+    });
+    chrome.tabs.query({active: true}, function (aTab) {
+      chillOut(aTab);
+    });
+  }
+});
+
+chrome.alarms.create('chillOut', {periodInMinutes: 1});
+if (chrome.pageAction) {
```
wbamberg: Why test for some APIs and not others? I'm not sure what the best practice is here really, but I think I'd prefer to just assume the APIs exist, given a sensible min-version value for Firefox.

ui-playground/lib/background.js
```diff
@@ -0,0 +1,114 @@
...
+  chrome.tabs.query({active: true}, tabs => {
+    tabs.forEach(tab => {
+      if (tab && tab.id) {
+        chrome.pageAction.show(tab.id);
+      }
+    });
+  });
+
+  chrome.tabs.onActivated.addListener(activeInfo => {
+    chrome.pageAction.show(activeInfo.tabId);
```
wbamberg: It seems like a shame to show the page action for every active tab, as the point of a page action is that, unlike a browser action, it's not always visible. Like this, we might wonder how page action differs from browser action.

(One possibility could be that when the alarm is triggered, the page action is shown for the active tab. Then if it is clicked, it calls chillout.)

ui-playground/lib/notify.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+/*eslint-env browser */
+
+window.postMessage('whimsy:enabled', '*');
```
wbamberg: I'm not sure what this content script is for.

ui-playground/manifest.json
```diff
@@ -0,0 +1,73 @@
...
+    "128": "images/whimsycorn-128.png"
+  },
+  "applications": {
+    "gecko": {
+      "id": "ui-playground@mozilla",
+      "strict_min_version": "40.0.0"
+    }
+  },
+  "permissions": [
+    "tabs",
```
wbamberg: I was surprised you needed "tabs" for this. It seems like in Firefox you need it in order to query by URL, or you get a failure here: https://dxr.mozilla.org/mozilla-central/source/browser/components/extensions/ext-tabs.js#431 because tab.url is undefined. But in Chrome you don't. I'll file a bug for this.

bwinton:
Hmmm…  Maybe this kind of add-on isn't really suitable for this repo, since it's not an example, as much as a playground/demonstration sandbox.

wbamberg:
OK. I think it could work, but would need more explanation around what all the bits of it do. But I'll close this PR for now anyway.

PR #24, opened 2016-01-05
improve link detection in notify-link-clicks

minj:
Original approach is a bit too naive and misses all links
that contain other elements which can grab mouse clicks.
E. g. simple target checking fails on the grid @ mozilla.org.

Traversing up the tree to look for valid links fixes this problem.


minj:
@wbamberg rebased the changes

wbamberg:
Thanks @minj !

PR #25, opened 2016-01-05
fix review comments for user-agent-rewriter

wbamberg:
@minj: I've tried to address your comments on the user-agent-rewriter example here. Please let me know if they look good to you.

### DIFF
--- a/user-agent-rewriter/popup/choose_ua.css
+++ b/user-agent-rewriter/popup/choose_ua.css
@@ -1,14 +1,19 @@
-html, body {
+html, body, .ua-choices {
   height: 100px;
   width: 100px;
   margin: 0;
 }
 
+.ua-choices {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-around;
+}
+
 .ua-choice {
   height: 20%;
-  width: 90%;
-  margin: 3% auto;
-  padding: 8% 6% 0 6%;
+  margin: 0.2em;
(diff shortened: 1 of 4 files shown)


user-agent-rewriter/popup/choose_ua.css
```diff
@@ -1,14 +1,19 @@
-html, body {
+html, body, .ua-choices {
   height: 100px;
   width: 100px;
```
minj: The default font on my system (manjaro Linux) resolves to ` "DejaVu Serif"`. It's a bit too wide:

102px seems to be the limit.

However, the popup works just fine with no `width` rule.

minj:
Otherwise, looks good to me :+1:

wbamberg:
Removing "width" does not work for me (OS X) as the popup is also too narrow:

overflow

120px works though, and it sounds like that works on your system too.

When I first wrote this example, Firefox didn't attempt to size the popup to fit the content at all (you can still see this behavior in Firefox 43). Now I think it does (https://dxr.mozilla.org/mozilla-central/source/browser/components/extensions/ext-utils.js#180), but I wonder if it's not working quite right...

minj:
Just checking, did you test the current nightly OR?..

wbamberg:
Yes, latest Nightly. It seems that for now, setting an explicit width is the recommended way to go here.

PR #26, opened 2016-01-06
cross browser workaround for chrome.tabs.executeScripts and minor style tweaks

rpl:
The current workaround applied on chrome.tabs.executeScript does not work on Chromium-based browsers.

This PR use an explicit absolute url which works on both Firefox and Chrome (and some minor tweaks on the stylesheet applied to the popup page)

### DIFF
--- a/beastify/popup/choose_beast.css
+++ b/beastify/popup/choose_beast.css
@@ -1,14 +1,10 @@
 html, body {
-  height: 100px;
   width: 100px;
-  margin: 0;
 }
 
 .beast {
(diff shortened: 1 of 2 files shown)


rpl:
@wbamberg I've just tried the "beastify" example on Iridium (a Chrome based browser) and it doesn't work because of the different url resolution.

How about using an explicit absolute url which works on both Firefox and Chrome?

wbamberg:
Works great, thanks @rpl!

PR #34, opened 2016-01-30
add _locales/ja

lv7777:
add _locales/ja and modify messages.json.

this addon is not run in less than firefox46.
see https://bugzilla.mozilla.org/show_bug.cgi?id=1240354


wbamberg:
Thanks @lv7777 !

PR #37, opened 2016-02-13
a tabs example

andymckay:


tabs-tabs-tabs/manifest.json
```diff
@@ -0,0 +1,17 @@
...
+    "gecko": {
+      "id": "tabs-tabs-tabs@mozilla.org"
+    }
+  },
+  "browser_action": {
+      "default_title": "Tabs, tabs, tabs",
+      "default_popup": "tabs.html"
+  },
+  "permissions": [
+      "tabs"
```
wbamberg: Trailing comma here stops this from installing. You don't need the "tabs" permission for the operations in this extension, so I think we should just remove this.

tabs-tabs-tabs/manifest.json
```diff
@@ -0,0 +1,17 @@
...
+  ]
+}
```
wbamberg: In the other examples, we include manifest keys that are optional but are still good practice to include:
   * description
   * icons
   * homepage_url
   * strict_min_version in the applications key
   * default_icon in the browser_action key

tabs-tabs-tabs/tabs.html
```diff
@@ -0,0 +1,17 @@
...
+<body>
+  <a href="#" id="tabs-move-beginning">Move active tab to the beginning of the window</a><br>
+  <a href="#" id="tabs-move-end">Move active tab to the end of the window</a><br>
+  <a href="#" id="tabs-duplicate">Duplicate active tab</a><br>
+  <a href="#" id="tabs-reload">Reload active tab</a><br>
+  <a href="#" id="tabs-remove">Remove active tab</a><br>
+  <script src="tabs.js"></script>
+</body>
+
+</html>
```
wbamberg: This popup looks comically narrow on Chrome: https://i.imgur.com/mXfSNFm.png

tabs-tabs-tabs/tabs.js
```diff
@@ -0,0 +1,42 @@
...
+    });
+  }
+
+  else if (e.target.id === "tabs-remove") {
+    callOnActiveTab((tab) => {
+      chrome.tabs.remove(tab.id);
+    });
+  }
+
+});
```
wbamberg: This code doesn't work in Chrome. I get this in the console:

    Uncaught SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode

tabs-tabs-tabs/tabs.js
```diff
@@ -0,0 +1,42 @@
...
+
+  if (e.target.id === "tabs-move-end") {
+    callOnActiveTab((tab) => {
+      chrome.tabs.move([tab.id], {index: -1});
+    });
+  }
+
+  else if (e.target.id === "tabs-duplicate") {
+    callOnActiveTab((tab) => {
+      chrome.tabs.duplicate(tab.id);
```
wbamberg: This doesn't work in Firefox, because it's not implemented yet. I think we should either (1) not include stuff that's not implemented in Firefox or (2) test for such stuff and log a helpful error.

I think I'd vote for (1). We can add it back when support for duplicate() lands.

wbamberg:
This is awesome Andy. I was thinking about writing something quite similar actually :).

tabs-tabs-tabs/manifest.json
andymckay: Got that, thanks.

tabs-tabs-tabs/tabs.html
andymckay: Yeah, I'll see if i can improve that.

tabs-tabs-tabs/tabs.js
andymckay: Well its just about to https://bugzilla.mozilla.org/show_bug.cgi?id=1238312, I wrote this to help test the API.

tabs-tabs-tabs/tabs.js
andymckay: Yeah, got that fixed, thanks.

tabs-tabs-tabs/manifest.json
andymckay: Setting strict_min_version is probably good.

andymckay:
screenshot 2016-02-15 11 21 37

tabs-tabs-tabs/manifest.json
```diff
@@ -0,0 +1,17 @@
+{
+  "applications": {
+    "gecko": {
+      "id": "tabs-tabs-tabs@mozilla.org",
+      "strict_min_version": "47.0.0"
```
wbamberg: This doesn't install in my Nightly, I guess because it does not match 47.0a1. "47.*" doesn't, either (I guess "*" is interpreted as numeric (?)).

"47.0a1" works. Would it be OK to use that?

andymckay: Hmm, that's wierd, all the other addons in this repo have versions like "45.0.0" so I followed that.

wbamberg: Yes, but I'd guess that it thinks (correctly) that 45.0.0 < 47.0a1 < 47.0.0.

So if Nightly is 47.0a1, then 45.0.0 installs fine on Nightly, but 47.0.0 does not. But this add-on can't use 45.0.0, because it needs tabs.duplicate().

"47.0a1" will work for now, and I think it will still work when 47 is 47.0. If not we can change it then.

This makes me think we need better docs for version numbers...

PR #42, opened 2016-02-21
Add an example for using the bookmarks API

johannhof:
We just landed `bookmarks.search`, and I figured it's time to make an example of how to use the bookmarks API. I ended up also using a lot of the `browserAction` and `tabs` APIs, because I wanted it to have at least some basic "functionality".

The add-on itself works exactly like the bookmark button that is built into Firefox (and Chrome). It shows if the current page is bookmarked and toggles a bookmark for the current page on click.


bookmark-it/manifest.json
```diff
@@ -0,0 +1,36 @@
...
+    "96": "icons/bookmark-it@2x.png"
+  },
+
+  "manifest_version": 2,
+  "name": "bookmark-it",
+  "version": "1.0",
+
+  "applications": {
+    "gecko": {
+      "id": "bookmark-it-extension@mozilla.org"
```
wbamberg: This should have strict_min_version. If search() only exists in Nightlies, I think we could set it to "47.0a1".

bookmark-it/manifest.json
```diff
@@ -0,0 +1,36 @@
...
+  "name": "Bookmark it!",
+  "version": "1.0",
+  "description": "A simple bookmark button",
+  "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/bookmark-it",
+  "icons": {
+    "48": "icons/bookmark-it.png",
+    "96": "icons/bookmark-it@2x.png"
+  },
+
+  "manifest_version": 2,
```
wbamberg: This key appears twice.

bookmark-it/manifest.json
```diff
@@ -0,0 +1,36 @@
...
+  "name": "bookmark-it",
+  "version": "1.0",
```
wbamberg: Oh hang on, so do these ones :)

johannhof: What did I do? :smile:

bookmark-it/background.js
```diff
@@ -0,0 +1,64 @@
...
+    },
+    tabId: currentTab.id
+  });
+}
+
+/*
+ * Add or remove the bookmark on the current page.
+ */
+function toggleBookmark() {
+  if(currentBookmark) {
```
wbamberg: nit: I usually see a space between "if" and "(condition)":

    if (condition)

...there's another one of these at line 44.

wbamberg:
Looks great, thanks @johannhof ! Just the few comments above.

johannhof:
Thanks, I'll fix!

bookmark-it/manifest.json
johannhof: Should I leave a notice in the README that it will only work with the latest nightly or will that be outdated too quickly anyway?

wbamberg: Good question. Maybe add a notice that it will work only from Firefox 47 onwards, then it will not get outdated?

johannhof:
@wbamberg I amended my previous commit. :)

bookmark-it/manifest.json
johannhof: Sounds like a solution!

PR #44, opened 2016-03-03
Change chrome APIs to browser.

jonathanKingston:
See the same change here: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Your_second_WebExtension$revision/1026038

Thanks

### DIFF
--- a/beastify/content_scripts/beastify.js
+++ b/beastify/content_scripts/beastify.js
@@ -7,7 +7,7 @@ beastify():
 function beastify(request, sender, sendResponse) {
   removeEverything();
   insertBeast(request.beastURL);
-  chrome.runtime.onMessage.removeListener(beastify);
+  browser.runtime.onMessage.removeListener(beastify);
 }
 
 /*
@@ -34,7 +34,7 @@ function insertBeast(beastURL) {
 /*
 Assign beastify() as a listener for messages from the extension.
 */
-chrome.runtime.onMessage.addListener(beastify);
+browser.runtime.onMessage.addListener(beastify);
 
 
 /*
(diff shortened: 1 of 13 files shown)


johannhof:
This means we could also use the new Promise API, would probably make for much nicer code!

jonathanKingston:
@johannhof I can't find docs for that unless you have some. I have seen some mention of the downloads API having promises.

andymckay:
Is there an advantage to this? I can only see the disadvantage that all the APIs won't work across browsers?

johannhof:
@jonathanKingston actually all `browser` APIs should now support returning Promises (if they're async) :)

Anyway, I agree with @andymckay that unless there is some great advantage to this it wouldn't be practical to get rid of the cross-browser compatibility.

So far the only Firefox-exclusive feature that comes with `browser.*` is Promises support, afaik. And I'm not sure if that's worth the switch.

andymckay:
Is anything documenting the difference, for example on this page? https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Chrome_incompatibilities

Its also worth nothing that all the MDN docs use chrome.* so they at the very least would be incompatible with this change.

jonathanKingston:
@andymckay all the API docs are using browser (this is what sparked me to make the change) for example:
- https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/executeScript
- https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/bookmarks/move

@johannhof Yup I haven't seen any of the mention of promise support until I read the download api after you mentioned it.

The main advantage would be consistency and explaining they are separate. If however cross compatibility is more important there is probably a decent way to explain the APIs are available under both `chrome` and `browser` global objects.

There probably would be some worth in making a simple shim to make cross compatibility work instead perhaps?

andymckay:
@jonathanKingston how did I miss that... sorry, you are right.

andymckay:
All the docs have now changed to using `chrome`.

kumar303:
I think cross-compatibility is indeed a goal. Anyway, it sounds like this was mostly a documentation problem that has been fixed. Is it ok to close this?

wbamberg:
Yes, I'm happy to close this. I think the chrome/browser question is still not finally resolved, but any resolution probably won't happen in this PR.

PR #45, opened 2016-03-07
Add an example for using the commands API

matthewjwein:


commands/background.js
```diff
@@ -0,0 +1,9 @@
+browser.commands.getAll(function(commands) {
```
wbamberg: This won't work in Chrome because it uses "browser", not "chrome". I think, if we can, we should be cross-browser here.

commands/manifest.json
```diff
@@ -0,0 +1,21 @@
...
+    "scripts": ["background.js"]
+  },
+  "applications": {
+    "gecko": {
+      "id": "commands@mozilla.org"
+    }
+  },
+  "commands": {
+    "toggle-feature": {
+      "suggested_key": { "default": "Ctrl+Shift+Y" },
```
wbamberg: This also won't work in Chrome, because Chrome has stricter controls on what you can use (Ctrl+Shift+5 works in Chrome)

commands/manifest.json
```diff
@@ -0,0 +1,21 @@
...
+  "description": "Press Ctrl+Shift+Y to send an event (Command+Shift+Y on a Mac).",
+  "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/commands",
+  "manifest_version": 2,
+  "version": "1.0",
+  "background": {
+    "scripts": ["background.js"]
+  },
+  "applications": {
+    "gecko": {
+      "id": "commands@mozilla.org"
```
wbamberg: I think we should include strict_min_version in here, especially for APIs that require newer Firefoxes:

"strict_min_version": "48.0a1"

...should work.

commands/background.js
```diff
@@ -0,0 +1,9 @@
...
+  commands.forEach(function(command) {
```
wbamberg: It would be nice to have brief comments for these 2 functions.

wbamberg:
Thanks @matthewjwein ! I just had a few comments.

commands/manifest.json
matthewjwein: Hm, I mainly chose this shortcut because the chrome example in https://developer.chrome.com/extensions/commands uses "Ctrl+Shift+Y", and so does their sample extension: https://developer.chrome.com/extensions/samples#sample-extension-commands-extension.

matthewjwein: I just tested Chrome's sample extension that uses "Ctrl+Shift+Y" and noticed that it doesn't work on Chrome either - it did work when I switched to "Ctrl+Shift+5".   Do you know why that is? I chose the shortcut by referencing the API page, which says that the supported keys are: A-Z, 0-9, Comma, Period, Home, End, PageUp, PageDown, Space, Insert, Delete, Arrow keys (Up, Down, Left, Right) and the Media Keys (MediaNextTrack, MediaPlayPause, MediaPrevTrack, MediaStop).

wbamberg: Yes, sorry, I misread the docs, and your code should work in Chrome. I found this: https://stackoverflow.com/questions/19599717/chrome-commands-api, which helped.

matthewjwein: Thanks, that worked.   The shortcut "Ctrl+Shift+Y" does work for me.  I just needed to uninstall and reinstall the extension.

commands/manifest.json
matthewjwein: Done.

matthewjwein:
Thanks for the review!

wbamberg:
Thanks @matthewjwein !

PR #46, opened 2016-03-08
an example using page action and downloads.download

andymckay:


download-xpi/background.js
```diff
@@ -0,0 +1,20 @@
+var rx = new RegExp('^https://addons.mozilla.org/[A-Za-z\-]+/[a-z]+/addon/');
+var xpi = new RegExp('^https://addons.mozilla.org/[a-z]+/downloads/file/[0-9]+/');
+
+function download(url) {
+  if (url.match(xpi)) {
+    chrome.downloads.download({'url': url});
+  } else {
+    console.log('Not a valid XPI, ignoring:', xpi)
```
wbamberg: If I visit addons.mozilla.org, and click on some add-ons, I see the page action but when I click it I get "Not a valid XPI" logged to the console. The URL is something like "https://addons.mozilla.org/firefox/downloads/latest/2592/addon-2592-latest.xpi?src=hp-dl-featured" (note "latest" instead of "file"). I think it would be good to either tweak the regexp, or use a more constrained scenario.

download-xpi/background.js
```diff
@@ -0,0 +1,20 @@
+var rx = new RegExp('^https://addons.mozilla.org/[A-Za-z\-]+/[a-z]+/addon/');
```
wbamberg: This variable name could be more meaningful.

wbamberg: I'd like to have some brief comments in this file.

wbamberg:
Looks great, just a few comments. I couldn't get it to work on Chrome though, not sure why...

andymckay:
Works in Chrome now, thinking this add-on is a bit finnicky and maybe more complicated than useful though.

PR #49, opened 2016-03-22
handle pinned tabs correctly

wbamberg:
tabs.move() seems to treat pinned and unpinned tabs groups independently: so you can't move an unpinned tab before a pinned tab, or a pinned tab after an unpinned tab.

For example:
* if you have one pinned tab, then you can't move an unpinned tab to index 0
* if you have one unpinned tab, then you can't move a pinned tab to index -1

If you try, you get a silent failure.

This patch is intended to respect that behavior:

**Move active tab to the beginning of the window**:

* if active tab is pinned, move to index 0
* if active tab is unpinned, move to the first unpinned index

**Move active tab to the end of the window**:

* if active tab is pinned, move to the first unpinned index
* if active tab is unpinned, move to -1

### DIFF
--- a/tabs-tabs-tabs/tabs.js
+++ b/tabs-tabs-tabs/tabs.js
@@ -8,7 +8,7 @@ function firstUnpinnedTab(tabs) {
 
 document.addEventListener("click", function(e) {
   function callOnActiveTab(callback) {
-    chrome.tabs.query({}, function(tabs) {
+    chrome.tabs.query({windowId: chrome.windows.WINDOW_ID_CURRENT}, function(tabs) {
       for (var tab of tabs) {
         if (tab.active) {
           callback(tab, tabs);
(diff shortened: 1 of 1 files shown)


tabs-tabs-tabs/tabs.js
```diff
@@ -1,23 +1,39 @@
+function firstUnpinnedTab(tabs) {
+  for (var tab of tabs) {
```
kmaglione: This will include tabs in all windows, so it's not going to give you what you expect when multiple windows are open.

tabs-tabs-tabs/tabs.js
```diff
@@ -1,23 +1,39 @@
...
+  }
+}
+
 document.addEventListener("click", function(e) {
   function callOnActiveTab(callback) {
     chrome.tabs.query({}, function(tabs) {
       for (var tab of tabs) {
         if (tab.active) {
-          callback(tab);
+          callback(tab, tabs);
```
kmaglione: Unless you're intentionally calling the callback once for every window, you should return here.

tabs-tabs-tabs/tabs.js
```diff
@@ -1,23 +1,39 @@
...
         }
       }
     });
   }
 
   if (e.target.id === "tabs-move-beginning") {
-    callOnActiveTab((tab) => {
-      chrome.tabs.move([tab.id], {index: 0});
+    callOnActiveTab((tab, tabs) => {
+      var destination = 0;
```
kmaglione: If you call this `index` rather than `destination`, you can use ES6 shorthand below and get `{index}` rather than `{index: destination}`

tabs-tabs-tabs/tabs.js
wbamberg: I've tried to restrict this to tabs in the current window, to fix this problem and the one above. This seems to work OK with multiple windows.

tabs-tabs-tabs/tabs.js
wbamberg: Thanks! I did not know this.

tabs-tabs-tabs/tabs.js
```diff
@@ -1,23 +1,40 @@
...
+    if (!tab.pinned) {
+      return tab.index;
+    }
+  }
+}
+
 document.addEventListener("click", function(e) {
   function callOnActiveTab(callback) {
-    chrome.tabs.query({}, function(tabs) {
+    chrome.tabs.query({windowId: chrome.windows.WINDOW_ID_CURRENT}, function(tabs) {
```
kmaglione: You'd be better off using `currentWindow: true`. Firefox only supports `WINDOW_ID_CURRENT` as of 48.

kmaglione:
r+wc

Thanks!

PR #60, opened 2016-05-17
Use browser styles…

bwinton:
What do you think of this, Will?  🙂

### DIFF
--- a/latest-download/manifest.json
+++ b/latest-download/manifest.json
@@ -19,9 +19,10 @@
   "permissions": [
     "downloads",
     "downloads.open"
-  ],  
+  ],
 
   "browser_action": {
+    "browser_style": true,
     "default_icon": "icons/page-32.png",
     "default_title": "Latest download",
     "default_popup": "popup/latest_download.html"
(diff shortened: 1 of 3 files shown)


bwinton:
Empty:
screenshot 2016-05-17 14 52 36

Populated:
screenshot 2016-05-17 14 53 40

wbamberg:
That is lovely! I should nag you more often.

bwinton:
Dammit, once again I fell for the “positive reinforcement of negative behaviour” trap!  😄

wbamberg:
The only problem I just noticed is that disabling the buttons doesn't work any more: https://github.com/mdn/webextensions-examples/blob/master/latest-download/popup/latest_download.js#L37, I guess because they are no longer input elements.

Do you know if there's a simple fix for this using browser_styles?

bwinton:
Hmm, try adding the "disabled" class to the buttons instead?  (And check for it in the event handlers, of course…)
I think something like `document.querySelector("#open").classList.add("disabled");` should work, but I'm not totally sure.  @shorlander might have more better ideas.

wbamberg:
This works to stop the popup from trying to open/remove the item, but the nice thing about input.disabled (apart from being more semantic) is that the buttons are made to appear disabled. Do you know if that's possible using browser styles?

bwinton:
Yeah, it looks like we don't have any browser styles for disabled footer buttons, which is mostly why I was looping in @shorlander…  :wink:

Issue #40, opened 2016-02-19
tabs-tabs-tabs needs a review


wbamberg:
https://github.com/mdn/webextensions-examples/tree/master/tabs-tabs-tabs

kmaglione:
```js
  function callOnActiveTab(callback) {
    chrome.tabs.query({active: true, currentWindow: true}, (tabs) => {
      tabs.forEach(callback);
    });
  }
```

Your click listener also needs an `e.preventDefault()`

Otherwise, looks good.

lv7777:
I tried this extension.
I suggest add feature that  "move new window this tab" and "alert info this tab"

wbamberg:
=> https://github.com/mdn/webextensions-examples/commit/8cce7fda06ac8d57bd09fe6caea150825d90dc63

wbamberg:
> I suggest add feature that "move new window this tab" and "alert info this tab"

These are good suggestions, and I'd happily take PRs for them, but I'm not planning to add them myself right now.

lv7777:
ok.
Actually, I implemented this feature for learning.
I may send pull request.

PR #67, opened 2016-06-04
create new feature.(alert info)

lv7777:
#40 
in tabs-tabs-tabs.

### DIFF
--- a/tabs-tabs-tabs/tabs.js
+++ b/tabs-tabs-tabs/tabs.js
@@ -60,5 +60,15 @@ document.addEventListener("click", function(e) {
     chrome.tabs.create({url: "https://developer.mozilla.org/en-US/Add-ons/WebExtensions"});
   }
 
+  else if (e.target.id === "tabs-alertinfo") {
+    callOnActiveTab((tab) => {
+      let props = "";
+      for (let item in tab) {
+        props += `${ item } = ${ tab[item] } \n`;
+      }
+      alert(props);
+    });
+  }
+
   e.preventDefault();
 });
--- a/tabs-tabs-tabs/tabs.html
+++ b/tabs-tabs-tabs/tabs.html
@@ -13,6 +13,7 @@
   <a href="#" id="tabs-reload">Reload active tab</a><br>
   <a href="#" id="tabs-remove">Remove active tab</a><br>
   <a href="#" id="tabs-create">Create a tab</a><br>
+  <a href="#" id="tabs-alertinfo">Alert active tab info</a><br>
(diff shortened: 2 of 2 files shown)


tabs-tabs-tabs/tabs.js
```diff
@@ -55,5 +55,14 @@ document.addEventListener("click", function(e) {
       chrome.tabs.remove(tab.id);
     });
   }
-
+  
+  else if (e.target.id === "tabs-alertinfo") {
+    callOnActiveTab((tab) => {
+      let props="";
```
wbamberg: Please use spaces between operators and operands, like:
`let props = "";`

tabs-tabs-tabs/tabs.js
```diff
@@ -55,5 +55,14 @@ document.addEventListener("click", function(e) {
...
+      for(let a in tab){
```
wbamberg: Again:
`for (let a in tab) {`

tabs-tabs-tabs/tabs.js
```diff
@@ -55,5 +55,14 @@ document.addEventListener("click", function(e) {
...
+        props+= a+"="+tab[a]+"\n"
+      }
+      alert(props)
```
wbamberg: Semicolon here please:
 `alert(props);`

wbamberg:
Thanks a lot for adding this. I had a few formatting comments, but it seems to work fine on Firefox.

Weirdly, on Chrome, the alert box appears then disappears again immediately. I think this must be because the browser action popup closes as soon as you click the link, and when that closes, its alert is closed as well. If you keep the popup open (by selecting "inspect popup"), then the alert stays open, too.

But then, the popup closes in Firefox too, and the alert stays open there. So maybe that is just a difference in behavior.

I think it would be good for these examples to work in Chrome where that's possible, so it would be good to find a solution for this. We could just log the info to the console (easy), or display it inside the popup (hard). Or maybe use the alert() in the currently active tab, instead of the one in the popup?

wbamberg:
Also, you need to update your branch from master, as it doesn't apply cleanly at the moment (it's missing https://github.com/mdn/webextensions-examples/pull/62/files and https://github.com/mdn/webextensions-examples/commit/8cce7fda06ac8d57bd09fe6caea150825d90dc63).

lv7777:
Thanks for advice and sorry for late.

I just fix formatting problem,and will sending pull request again.

lv7777:
sended.

wbamberg:
Thanks @lv7777 !

PR #68, opened 2016-06-07
provide an example using the history api

andymckay:


history-deleter/history.js
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// When the page is loaded find the current tab and then use that to query
+// the history.
+chrome.tabs.query({active: true, currentWindow: true}, function(tabs) {
+  var list = document.getElementById('history');
+  var hostname = get_hostname(tabs[0].url);
+  chrome.history.search(
+    // Search for all history entries for the current windows domain.
+    {text: hostname},
```
bobsilverberg: You can also pass `maxResults: 5` to limit the search to 5 results, rather than limiting to 5 in the loop below.

history-deleter/history.js
```diff
@@ -0,0 +1,58 @@
+// A useful way to extract the domain from a url.
+function get_hostname(url) {
+  var a = document.createElement('a');
+  a.href = url;
+  return a.hostname;
+}
+
+function no_history(hostname) {
+  document.getElementById('history').innerHTML = `No history for ${hostname}.`;
```
bobsilverberg: This is kind of duplicating what's on line 15. Instead of this you could simply inline this in the callback and set the `innerHTML` of `list` in there. Of course, this is a demo for teaching purposes, so if you feel it's clearer to leave this as a separate function that's valid too.

history-deleter/history.js
```diff
@@ -0,0 +1,58 @@
...
+function clearAll(e) {
+  chrome.tabs.query({active: true, currentWindow: true}, function(tabs) {
+    var hostname = get_hostname(tabs[0].url);
+
+    chrome.history.search(
+      {text: hostname},
+      // Search will return us a list of histories for this domain.
+      // Loop through them and delete them one by one.
+      function(results) {
+        for (k = 0; k < results.length; k++) {
```
bobsilverberg: I'd probably use a `for...of` loop here, but that's just a preference.

history-deleter/history.js
```diff
@@ -0,0 +1,58 @@
...
+          chrome.history.deleteUrl({url: results[k].url});
+        }
+        // Clear out the UI.
+        no_history(hostname);
```
bobsilverberg: Oh, now I see why `no_history`. :)

history-deleter/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+The popup shows a list of 5 history entries for the current domain. It provides a clear button to delete all entries for that domain.
+
+## What it shows
+
+How to use the history API.
+
+## Caveats
+
+When using it from about:* pages (for example about:debugging) you have no hostname so that does a history query for everything.
```
bobsilverberg: It might be worth pointing out here, or in a code comment, that this also means that clicking the clear link will also delete *everything*.

history-deleter/history.js
andymckay: Ah I did not know that, thanks.

history-deleter/README.md
andymckay: Perhaps I should stop that?

andymckay:
Updated with feedback from @bobsilverberg and made it page_action to stop it showing on about: pages where it would be bad.

wbamberg:
This looks great to me, and works fine in Firefox. It does not seem to work very reliably in Chrome.... after clearing history for a domain I still see the history entries appear under the "History" menu, but when I click "Show Full History" I do not see the history entries: https://i.imgur.com/dGq711z.gif.

But I think this is a Chrome bug, and I'm happy to merge this if @bobsilverberg is.

@andymckay , do you want collaborator access, or whatever they call it, for you or anyone else in the team, so you can merge PRs? It seems silly for me to be a bottleneck in cases like this, where a PR's already had review.

andymckay:
Sorry for the late reply, I'm fine not having commit access.

I gave the STR a try in Chrome and couldn't reproduce. My guess is that its probably a timing issue between clearing and re-querying caching on Chrome?

PR #79, opened 2016-07-21
remove window.parent check from contentscript

groovecoder:
The manifest 'all_frames' property defaults to false, so the contentscript should not need to perform the window.parent check.

See https://github.com/mozilla/blok/pull/29#discussion_r71736456

### DIFF
--- a/inpage-toolbar-ui/contentscript.js
+++ b/inpage-toolbar-ui/contentscript.js
@@ -22,15 +22,13 @@ function toggleToolbar(toolbarUI) {
   }
 }
 
-// Handle messages from the add-on background page (only in top level iframes)
-if (window.parent == window) {
-  chrome.runtime.onMessage.addListener(function(msg) {
-    if (msg == "toggle-in-page-toolbar") {
-      if (toolbarUI) {
-        toggleToolbar(toolbarUI);
-      } else {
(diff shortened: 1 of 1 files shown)


wbamberg:
Looks good to me. @rpl , what do you think?

Also, @groovecoder , we're supposed to be adding some scary warnings to the docs for this, telling people about the risks of hostile web pages obscuring/replacing the injected content (https://bugzilla.mozilla.org/show_bug.cgi?id=1287590).

rpl:
@wbamberg I'm trying to remember if the check is there because "all_frames" wasn't implemented at the time this example has been added to this repo:

Given that this example is currently marked as compatible with Firefox >= 46, it can likely be the reason, in that case we should update the strict_min_version as well (to Firefox >= 48 or to the version Firefox >= 47 where "all_frames" landed, https://bugzilla.mozilla.org/show_bug.cgi?id=1213993)

(I also pointed out Bug 1287590 as part of the comment of that Blok PR, https://github.com/mozilla/blok/pull/29#issuecomment-234312358)

wbamberg:
We removed this example, because it was felt that it's not a good pattern to use: https://github.com/mdn/webextensions-examples/pull/87.

PR #85, opened 2016-08-17
New example, showing insertCSS and removeCSS

wbamberg:


apply-css/manifest.json
```diff
@@ -0,0 +1,28 @@
...
+
+  "background": {
+    "scripts": ["background.js"]
+  },
+
+  "page_action": {
+    "default_icon": "icons/off.svg"
+  },
+  
+  "permissions": [
```
Cobe1: (admin)

apply-css/README.md
```diff
@@ -0,0 +1,18 @@
+# apply-css
```
Cobe1: "apt replace css
apt install o' (Admin) cobe1" push C' p'

apply-css/manifest.json
Cobe1: (admin)

apt install o' (user: wbamburg) to read write and limited transfer maneuverability "permissions"

apply-css/manifest.json
```diff
@@ -0,0 +1,28 @@
...
+      "strict_min_version": "49.0"
+    }
+  },
+
+  "background": {
+    "scripts": ["background.js"]
+  },
+
+  "page_action": {
+    "default_icon": "icons/off.svg"
```
bobsilverberg: There is a message in the console when loading the add-on stating `WARN	Please specify whether you want browser_style or not in your page_action options.`. It doesn't break anything, but if it is something one is meant to specify then maybe it should be added to the example.

apply-css/background.js
```diff
@@ -0,0 +1,52 @@
...
+function initializePageAction(tabId) {
+  chrome.pageAction.setIcon({tabId, path: "icons/off.svg"});
+  chrome.pageAction.setTitle({tabId, title: TITLE_APPLY});
+  chrome.pageAction.show(tabId);
+}
+
+/*
+When first loaded, initialize the page action for all tabs.
+*/
+chrome.tabs.query({}, (tabs)=> {
```
bobsilverberg: Space needed between `(tabs)` and `=>`.

bobsilverberg:
Looks good and works well, @wbamberg. Nice work! This is a nice, simple example. Are you thinking of creating another example that shows some of the other options, such as loading css from a file and targeting a specific tab, or do you think that should be added to this, or perhaps just not covered?

wbamberg:
Thanks for the review! I also added a check that we only show the action in http(s) pages, since otherwise it gets shown in about: pages, in which it doesn't work.

> Are you thinking of creating another example that shows some of the other options, such as loading css from a file and targeting a specific tab, or do you think that should be added to this, or perhaps just not covered? 

I hadn't really thought about it. In general, I'm not against having more examples at all, and more coverage is always good I suppose.. However I don't think it's practical for me to cover every possible way to use all the APIs, and I feel as if having mega-examples that covers tons of options might get confusing.

PR #86, opened 2016-08-19
Add an example of how to use export helpers in content scripts

wbamberg:


Cobe1:
appreciate the guidance definitely could be more available heading to the
mountains for a brothers anniversary will integrate any needed pre task for
success this week any literature non binding would be greatly appreciated
in up to speed with environment thx

andymckay:
Looks like we need to get something to review this.

andymckay:
@rpl could you take a r? please?

export-helpers/README.md
```diff
@@ -0,0 +1,28 @@
...
+* execute a content script in the current tab, when the user clicks a browser action
+* listen for messages from the content script, and display a notification when the message arrives.
+ 
+The content script then:
+ 
+ * defines a function `notify()` and exports that to the page as a property
+ of the global `window` object.
+ * defines an object `messenger`, that has a member function `notify()`, and
+ exports that to the page as a property of the global `window` object.
+ * creates an object `messenger2` directly in the page's scope
```
rpl: it looks like `messenger2` is not defined (yet?)

export-helpers/README.md
```diff
@@ -0,0 +1,28 @@
...
+JavaScript objects defined in content scripts with scripts loaded by web pages.
+
+It has a background script that does two things:
+
+* execute a content script in the current tab, when the user clicks a browser action
+* listen for messages from the content script, and display a notification when the message arrives.
+ 
+The content script then:
+ 
+ * defines a function `notify()` and exports that to the page as a property
```
rpl: it could makes sense to explain that the first one  `exportFunction` and the second one `cloneInto with cloneFunctions option`.

rpl: Another use case of `cloneInto` (without the `cloneFunctions` option) that could be reasonable to provide in this example is the following:
- a content script has received from the messaging API (e.g. onMessage or a Port) and wants to forward it to the page into a custom event:
  - on Chrome the object properties are accessible to the webpage javascript code without doing anything
  - on Firefox the object properties are not accessible to the webpage javascript and the objects needs to be explicitly cloned using cloneInto before forwarding as the details of the custom event sent to the page.

The following commit shows a similar change applied to the WebVR Extension to make it to work correctly when running as a WebExtension on Firefox: 

- https://github.com/rpl/WebVR-Extension/commit/1548d4e40a2450fcd13ae8804415970bf9f7bd75

export-helpers/background.js
```diff
@@ -0,0 +1,30 @@
...
+  if (chrome.runtime.lastError) {
+    console.log(chrome.runtime.lastError);
+  }
+}
+
+/*
+Execute content script in the active tab.
+*/
+function loadContentScript() {
+  chrome.tabs.executeScript({
```
rpl: If the example is meant to be Firefox specific we could change all the `chrome.*` API calls into `browser.*` API calls
(and given that it demonstrates a Firefox specific feature, it looks like it could be, the only part that could be cross-browser is the suggested "forward an object received through the messaging API to the page in a custom event" scenario).

export-helpers/background.js
```diff
@@ -0,0 +1,30 @@
+function onExecuted() {
+  if (chrome.runtime.lastError) {
+    console.log(chrome.runtime.lastError);
+  }
+}
+
+/*
+Execute content script in the active tab.
```
rpl: Nit, we could indent the inline comment content to make them easier to be spotted.

export-helpers/README.md
wbamberg: If possible, I'd like to get this PR finished off, and we could look at this afterwards?

wbamberg:
@rpl , thanks for looking and I'm sorry to be so slow fixing it. Someone was asking about export helpers on IRC today so I thought I should get this finished off.

I fixed your review comments, except https://github.com/mdn/webextensions-examples/pull/86#discussion_r118972444 - if it's OK with you I would like to get this simple example merged now. We could file an issue for this extra case.

I changed things a bit: the content script is injected declaratively into https://github.com/mdn/webextensions-examples/tree/master/export-helpers, which doesn't exist yet, but it will. This should make it a bit easier to run.

rpl:
@wbamberg r+ 
Looks great (and I agree, we can add the CustomEvent scenario in a follow up).

PR #88, opened 2016-08-24
Change apply-css icon color to red so it is visible in Firefox Developer Edition

makyen:
The current icons are black.  That makes it invisible in the default Firefox Developer Edition theme. This changes the icons to red which is visible in both stock Firefox and Firefox Developer Edition. The red color also matches the border that this example applies to the page.


wbamberg:
Thanks @makyen !

PR #89, opened 2016-08-27
Added native messaging example

wbamberg:


wbamberg:
@aswan , would you mind taking a look at this for me? Is it too simple?

aswan:
Looks great, I think simple is good.
Just a few little suggestions:
- Include a note about opening the browser console to see the console message?
- The python script should be executable
- A few items specific to windows:  mention that they'll need to have python installed?  (I think thats safe to assume on linux and mac but perhaps not windows)  And they'll need to set a registry key properly (which will presumably be covered in the docs you'll link to).  And finally, windows doesn't handle scripts with the #! syntax (at least prior to all their recent linux emulation stuff, i don't know if it works there, but in any case that's quite new and probably not widely used yet).  in tests we've handled this by creating a .bat file that runs the python interpreter with the python script.  its clunky but it works.
- There are enough things that need to be just so that i suspect a lot of people will miss something so troubleshooting will be important.  That seems worthy of its own page or section of a page on mdn, but perhaps a link to it here?

wbamberg:
> Looks great, I think simple is good.
> Just a few little suggestions:
> 
> Include a note about opening the browser console to see the console message?

I've updated the README to be a bit more explicit here.

> The python script should be executable

You mean, the executable bits should be set? I think they are, no? I added a note in the README anyway.

> A few items specific to windows: mention that they'll need to have python installed? (I think thats safe to assume on linux and mac but perhaps not windows) And they'll need to set a registry key properly (which will presumably be covered in the docs you'll link to). And finally, windows doesn't handle scripts with the #! syntax (at least prior to all their recent linux emulation stuff, i don't know if it works there, but in any case that's quite new and probably not widely used yet). in tests we've handled this by creating a .bat file that runs the python interpreter with the python script. its clunky but it works.

OK, I've done these things, and it seems to work on Windows with this setup.

> There are enough things that need to be just so that i suspect a lot of people will miss something so troubleshooting will be important. That seems worthy of its own page or section of a page on mdn, but perhaps a link to it here?

It's a good idea, but I'd like to write the page before linking to it, since otherwise I'm certain to move it, or rename it, or something, and will just have to go back and update the link anyway.

I'm a bit torn on how much to automate build. The Chrome example automates everything, so you run an install script for your platform, and it takes care of all the fiddling with registry keys and file paths and stuff. This is really useful if you are actually building something, but for learning, I think it's helpful to go through the steps manually. I suppose ideally we should have a `build` directory as well, that automates it...

aswan:
I think that for the purposes of this example, having people do the setup themselves is completely reasonable (and that it is helpful for people who are going to go on to automate the installation process to just do it by hand once to see the steps).
Everything looks great, I think there's even a troubleshooting page now available for linking on MDN?

PR #92, opened 2016-09-15
add webNavigation API and webNavigation event filters example

rpl:


wbamberg:
Looks great @rpl ! Thanks for adding this. I'll do a little copy-editing to the README, but I think it's going to be easier to do that separately after merging.

PR #93, opened 2016-09-21
Add an example on how to use webpack with webextensions

freaktechnik:
This is a very basic example for how to use webpack to be able to use modules within a WebExtension. I am not sure if that's the desired configuration, but it is the simplest configuration that works. I am also not sure if there's some kind of coding style this repo tries to follow, as this loosely follows my personal style.

In ohter words, I am happy about feedback and improvement suggestions.


wbamberg:
Thanks @freaktechnik ! I'll take a look at this in the next couple of days.

webpack-modules/README.md
```diff
@@ -0,0 +1,31 @@
+# WebExtension Webpack Example
+A minimal example of how to use npm modules from within a WebExtension.
+The example package used by this extension is `left-pad`, an essential package
+in almost any situation.
+
+## What it does
+This setup takes a background and a content script entrypoint and outputs them
```
wbamberg: I think "left-pad.js" is not a content script is it? It's a popup script (meaning, it runs in the chrome process, can get direct access to the background page, and can access all the privileged APIs that the background page can).

This makes it a bit strange that the popup has the option of sending a message to the background script to do the padding. I think it would be better to omit the background script from this example, and just demonstrate using the popup script. Unless there's some reason I'm missing, that you are including both?

wbamberg: I think it would be good to split "what it does" in two parts here:
* what this example does: shows how to package npm modules for a WebExtension
* what the WebExtension does

webpack-modules/README.md
```diff
@@ -0,0 +1,31 @@
+# WebExtension Webpack Example
+A minimal example of how to use npm modules from within a WebExtension.
```
wbamberg: Some links would be good here, and a reference to webpack:

"A minimal example of how to use [webpack](https://webpack.github.io/) to package [npm](https://www.npmjs.com/) modules so they can be used in a WebExtension."

webpack-modules/README.md
```diff
@@ -0,0 +1,31 @@
...
+The example package used by this extension is `left-pad`, an essential package
+in almost any situation.
+
+## What it does
+This setup takes a background and a content script entrypoint and outputs them
+to their respective target.
```
wbamberg: I'm not at all familiar with webpack, so this might just be me, but it took me a little while to understand this. Can we be more explicit, like:

"This example shows how you can use webpack to to package npm modules for a WebExtension. The WebExtension's files are all stored under the "dist" directory. The WebExtension includes a browser action with a popup: the popup provides a UI for left-padding a string with a chosen character.

In this example, a file "popup/left-pad.js" implements the logic behind the popup's UI. This file uses the [left-pad](https://www.npmjs.com/package/left-pad) npm module. We can then use webpack to package this file as a self-contained JS file, which gets written to dist/popup/index.js, where it is included by the WebExtension's popup document."

webpack-modules/webpack.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+    entry: {
+        background_scripts: "./background_scripts/background.js",
+        popup: "./popup/left-pad.js"
+    },
+    output: {
+        path: "dist",
```
wbamberg: Is "dist" a convention for webpack output? I thought something like "addon" might be more obvious.

webpack-modules/README.md
freaktechnik: Left pad gets included once as background script and once as content script, thus two build targets. I did that to show, that you can use webpack both for content and background scripts.

webpack-modules/webpack.config.js
freaktechnik: This can be whatever, so sure, we can make it "addon". By the way, if it's not marked in the README yet, I should probably add that.

wbamberg:
Thanks for the updates, it looks great. One thing though - the initial add-on is now under /addon/dist, but webpack is exporting the built files directly under /addon. So the add-on won't find them. I think it would be better to abolish dist, and have:

/addon
    /icons
    /popup
    /manifest.json

wbamberg:
Thanks for the contribution @freaktechnik !

PR #94, opened 2016-09-22
added examples of sdk and bootstrapped add-ons transitioning to a webextension

rpl:
This pull request adds two new directories, embedded-webextension-bootstrapped and embedded-webextension-sdk, each of these directories contains 3 subdirectories:

- step0-legacy-addon: the original addon that wants to transition to a webextension (sdk and bootstrapped)
- step1-hybrid-addon: an hybrid version of the original addon from step0, which embeds a webextension (with the legacy part of the addon that provides just data migration code, the embedded webextension provides the new implementation, and imports the user data and preferences from the legacy part)
- step2-pure-webextension: once the user data has been migrated and all the feature are provided by the embedded webextension,  the legacy part can be removed and the hybrid addon can be turned in a simple webextension addon


embedded-webextension-bootstrapped/step1-hybrid-addon/webextension/background.js
```diff
@@ -0,0 +1,16 @@
+"use strict";
+
+browser.storage.local.get("super-important-user-setting")
```
wbamberg: Is there a reason we use "browser" here but "chrome" elsewhere in this file?

rpl: @wbamberg there's no reason, I was using the "chrome" namespace everywhere, but using the "browser" on the both sides (in the legacy part and the embedded webextension) looks nicer, but I missed to change a couple of "chrome" API methods call, I'm going to rename them to "browser" and update this PR.

embedded-webextension-sdk/step1-hybrid-addon/package.json
```diff
@@ -0,0 +1,21 @@
...
+      "title": "Super important user preference",
+      "type": "string",
+      "value": "saved superImportantUserPref value"
+    }
+  ],
+  "engines": {
+    "firefox": ">= 51.0a1",
+    "fennec": ">= 51.0a1"
+  },
+  "hasEmbeddedWebExtension": true,
```
wbamberg: trailing comma

embedded-webextension-sdk/step1-hybrid-addon/webextension/content-script.js
```diff
@@ -0,0 +1,4 @@
+browser.runtime.sendMessage({
+  type: "notify-attached-tab",
+  message: window.location.href,
```
wbamberg: trailing comma (seems to work fine though :)

embedded-webextension-sdk/step1-hybrid-addon/package.json
rpl: ouch! fixed ("never reorder properties in a json file without running a json validation on it")

embedded-webextension-sdk/step1-hybrid-addon/webextension/content-script.js
rpl: @wbamberg in a js files it is mostly considered a best practice (which prevent you from forgetting to add a comma when you add new fields to the object), the downside is that the "brain" sometimes start to add the trailing comma to json file automatically, which is annoying, because, on the contrary, makes it an invalid json :-(

wbamberg:
This looks great to me. I've taken a read through it, and also tested the upgrade path for both examples.

rpl:
@wbamberg great! :+1:

PR #96, opened 2016-09-23
updated README files for embedded webextension examples

wbamberg:
@rpl , I updated the README files for these examples. I'd appreciate a quick check that I've not mangled anything, if you have the time. After that I think we are done with this :). Thanks!


rpl:
@wbamberg 👍 both the updated README files looks great now, thanks!

r+

Issue #90, opened 2016-09-06
Webextension panel does not auto-resize according to the URLs length 


vasilicamihasca:
**Steps to reproduce:**
1.Install the following webextension as a temporary add-on:
[history-deleter.zip](https://github.com/mdn/webextensions-examples/files/456875/history-deleter.zip)
2.Navigate a bit through google images in order to gather some browsing history.

**Expected results:**
The entire content is correctly displayed in the panel.

**Actual results:**
When the URLs are too long the webextension panel does not auto-resize according to the links length and the last 5 visited sites are not entirely displayed.

**Additional notes:**
- This issue reproduces on Firefox 51.0a1 (2016-09-05) and Firefox 50.0a2 (2016-09-05) under Windows 10 64-bit.
- Please see the attached screenshot:
![2016-09-06_1701](https://cloud.githubusercontent.com/assets/18114210/18276563/114d87e6-7454-11e6-9c39-d4c49b4c2d17.png)

andymckay:
Yeah that's not great but I wanted to keep the add-on really simple and the focus of the add-on is on the history deleter API not on the length. Saying that if, someone wanted to send a really simple pull request that fixed this, that would be cool.

kartik95:
Fixed the bug. Pull request https://github.com/mdn/webextensions-examples/pull/101

kartik95:
Will my PR be merged anytime soon? Or does it need any changes? @andymckay 
I request you to please add 'hacktoberfest' label to this issue before merging. It would really help me. Thanks!

wbamberg:
Fixed by https://github.com/mdn/webextensions-examples/pull/101.

wbamberg:
@vasilicamihasca , could you verify the fix, please?

vasilicamihasca:
I am still able to reproduce this issue on Firefox 51.0a2 (2016-10-14) under Windows 10 64-bit.

![2016-10-14_1848](https://cloud.githubusercontent.com/assets/18114210/19394104/57b97028-9240-11e6-90f3-7799653f0df9.png)

I am not sure but it might be a regression here. Because I noticed that on Firefox 49.0.1 successfully appears a scrollbar and the URL’s are correctly displayed.

![2016-10-14_1853_001](https://cloud.githubusercontent.com/assets/18114210/19394113/5d09016a-9240-11e6-877b-1c57cfe6243f.png)

wbamberg:
[...]

PR #99, opened 2016-10-05
Bug fix for history-deleter panel not expanding

phoebeclarke:
A workaround for issue #90.

Rather than resizing the panel, store the URLs as clickable links. This way they're kept at a fixed length.

linklist

I also made a few other (very minor) CSS changes.

### DIFF
--- a/history-deleter/history.css
+++ b/history-deleter/history.css
@@ -1,4 +1,12 @@
 html, body {
     margin: 0.2em;
     width: 350px;
+    padding: 0 0.2em;
+}
+#history {
+    width: 100%;
+}
+.link {
+    width: 100%;
+    display:block;
 }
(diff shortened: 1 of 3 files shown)


wbamberg:
Thanks so much for contributing! But I think it's better to show the actual URL in the panel. I played with this a bit, and it looks as if CSS like this will fix the wrapping/resizing problem:

```
html, body {
  margin: 0.2em;
}

.history-entry {
  display:block;
  white-space: nowrap;
  margin: 1em;
}
```

...with "history-entry" applied to the \<p\> elements that we create. What do you think?

phoebeclarke:
That's a neat solution :) And better in terms of the user being able to quickly check their previous history; mine obscures the information.

andymckay:
It looks like this is being done in #101 ?

phoebeclarke:
Yeah, his solution is much better. I'll close my PR.

PR #107, opened 2016-10-13
Link to what a page action is

kumar303:
I didn't know what a page action was so I had a hard time understanding how to test this example


wbamberg:
Thanks @kumar303 !

PR #108, opened 2016-10-13
Add link to page action in readme

kumar303:
Since I'm a n00b and didn't know that a page action is code for an icon in the URL bar, I didn't know how to test this example


wbamberg:
Thanks @kumar303 !

Issue #106, opened 2016-10-13
There is no way to unbeastify a web page


kumar303:
Steps to reproduce:
* Install the `beastify` extension
* Navigate to a web page
* Beastify the web page, such as selecting 'Frog'

In this state, there is no intuitive way to "unbeastify" the web page, i.e. return it to its original form. You can refresh the URL but that isn't exactly obvious.

I know this is just an example extension but showing some code to restore the original web page might also be a useful example for someone.

nistath:
I'm on it! I will have to remove the removeEverything method though, and replace it with some sort of overlay.

wbamberg:
Thanks @nistath ! I did wonder if you might also consider unbeastifying by just reloading the page. Just a suggestion though, feel free to try any approach that seems to work :).

wbamberg:
Actually, one other thing about this issue: we use this example in an introductory walkthrough doc: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Your_second_WebExtension, and for that reason I'm keen to keep the example quite simple. So if the fix for this is simple, that's great, but if it adds a lot of extra complexity, it might not be such a good idea.

Is the idea something like this: that if you click the button,
* if the page has not been beastified, then the popup looks as it does now
* if the page has been beastified, then the popup contains an "unbeastify" button?

If so, the tricky bit seems to me to be knowing whether the page has been beastified or not, at the time you display the popup.

nistath:
I've implemented 2 unbeastifying methods: one that removes everything and another that reloads.
I'll try to get them to show only if the page has been beastified.
I've also updated the documentation.
I'll PR as I work on it.

nistath:
I'll add the unbeastify/clear buttons via the javascript.

nistath:
I can't know whether or not the page has been beastified at the time the popup displays, without making the code more complicated than it needs to be for an introductory level.

I've written good documentation and I think this example has gotten as advanced as it should.

nistath:
#109 is pretty final right now. Can we merge it?

wbamberg:
Fixed by https://github.com/mdn/webextensions-examples/pull/109.

PR #109, opened 2016-10-13
Unbeastifying methods

nistath:
#106


beastify/popup/choose_beast.html
```diff
@@ -7,9 +7,11 @@
...
 
 <body>
-  <div class="beast">Frog</div>
-  <div class="beast">Turtle</div>
-  <div class="beast">Snake</div>
+  <div class="button beast">Frog</div>
+  <div class="button beast">Turtle</div>
+  <div class="button beast">Snake</div>
+  <div class="button clear inactive">Reload</div>
+  <div class="button clear inactive">Blank</div>
```
wbamberg: I don't think we should have 2 buttons. We would only need one, that restores the page to its pre-beastified state. "Reset" might be a better name than "reload".

beastify/popup/choose_beast.js
```diff
@@ -1,6 +1,30 @@
 /*
 Given the name of a beast, get the URL to the corresponding image.
 */
+
+/*
+updateButtons():
+* looks in the self.beast variable to see if the page is beastified or not,
```
wbamberg: This won't work, because the popup's document is reloaded each time the popup gets opened, so it'll be easy for it to get out of sync with the state of the page.

beastify/popup/choose_beast.css
```diff
@@ -2,15 +2,31 @@ html, body {
...
+.button:hover {
   background-color: #CFF2F2;
 }
+
+.beast {
+  background-color: #E5F2F2;
+}
+
+.clear {
+  background-color: #E74C3C;
```
wbamberg: It's nice for the button to be a different color from the others, but this is a bit bright. #FBFBC9 is a bit more subtle.

beastify/popup/choose_beast.js
```diff
@@ -12,33 +36,57 @@ function beastNameToURL(beastName) {
...
+    chrome.tabs.executeScript(null, {
+      file: "/content_scripts/clear.js"
+    });
 
-  chrome.tabs.executeScript(null, {
-    file: "/content_scripts/beastify.js"
-  });
+    chrome.tabs.query({active: true, currentWindow: true}, function(tabs) {
+      chrome.tabs.sendMessage(tabs[0].id, {type: chosenMethod});
+    });
```
wbamberg: If you use reloading to reset, you don't need to have a content script here. You can do it directly from the popup script using code like:

chrome.tabs.reload()

This will make things much easier!

wbamberg:
This seems a bit fragile - it's easy to get into a state where the popup stops working, if you press the buttons more than once.

I think what I'd suggest is:
* add just one extra button, labeled "Reset"
* show it all the time (i.e. don't worry about keeping track of the page state)
* have it reset by calling chrome.tabs.reload()

I think this will be simpler and more reliable.

nistath:
This still has problems. After a reload, the beastify buttons stop working.

nistath:
It should now work, although it doesn't directly address the issue.

nistath:
It works, by closing the popup. Can we merge this or do you need anything else?

beastify/popup/choose_beast.css
```diff
@@ -2,15 +2,31 @@ html, body {
...
   margin: 3% auto;
   padding: 4px;
   text-align: center;
   font-size: 1.5em;
-  background-color: #E5F2F2;
   cursor: pointer;
 }
 
-.beast:hover {
+.button:hover {
```
wbamberg: I think this should be "beast"? "button" works, because it gets overridden by "clear", but it seems like we really only want to apply this to beasts, not to all buttons.

beastify/popup/choose_beast.css
```diff
@@ -2,15 +2,31 @@ html, body {
...
+
+.clear {
+  background-color: #FBFBC9;
+}
+
+.clear:hover {
+  background-color: #EAEA9D;
+}
+
+.inactive {
```
wbamberg: I think we don't need inactive any more.

beastify/popup/choose_beast.js
```diff
@@ -26,19 +26,24 @@ Then get the active tab and send "beastify.js" a message
...
+
+    chrome.tabs.executeScript(null, {
+      file: "/content_scripts/beastify.js"
+    });
+
+    chrome.tabs.query({active: true, currentWindow: true}, function(tabs) {
+      chrome.tabs.sendMessage(tabs[0].id, {beastURL: chosenBeastURL});
+    });
+    
     return;
```
wbamberg: You don't need this return statement.

beastify/popup/choose_beast.js
```diff
@@ -26,19 +26,24 @@ Then get the active tab and send "beastify.js" a message
...
-
-  chrome.tabs.executeScript(null, {
-    file: "/content_scripts/beastify.js"
-  });
-
-  chrome.tabs.query({active: true, currentWindow: true}, function(tabs) {
-    chrome.tabs.sendMessage(tabs[0].id, {beastURL: chosenBeastURL});
-  });
-
+    return;
```
wbamberg: You don't need this return statement.

wbamberg:
Thanks @nistath , I think this looks great. I had a few little changes, and could you please update the README: https://github.com/mdn/webextensions-examples/blob/master/beastify/README.md.

Then I'll be able to merge this.

beastify/popup/choose_beast.js
nistath: I'm keeping it as good practice. If we can't use a switch case to handle this (which would use breaks), we should at least have the return statements at the ifs as some sort of equivalent.

wbamberg:
Thanks @nistath !

Issue #103, opened 2016-10-12
Example for word replacements


jvillalobos:
A common type of add-on we see being created, mostly for entertainment purposes, are add-ons that make word/phrase replacements, like [XKCD Substitutions](https://addons.mozilla.org/addon/xkcd-substitutions/). This would also make a good example for beginners that could be reused to create real add-ons.

Discussing it with my team, we came up with the idea of a word -> emoji substitution add-on. It could use a simple mapping (from http://emojipedia.org/ or similar) to replace a bunch of common words for emojis.

One minor caveat is that we should make sure the replacement script has decent performance, since it could end up being replicated multiple times.

atsay:
The plan would be to turn this example into a "make your first add-on" tutorial that we can provide as an activity on the mozilla.org/contribute page and Campus Clubs.

aterranova-bv:
This sounds awesome, I'm on it! 🎃

jvillalobos:
🏆

wbamberg:
Fixed: https://github.com/mdn/webextensions-examples/pull/110.

PR #110, opened 2016-10-16
Adds an Emoji Substitution example

aterranova-bv:
# Issue

Addresses #103.

# Description

This creates a new example called "Emoji Substitution" that replaces all occurrences of a bunch of common emoji words with their actual emoji counterparts.  For example, `pizza` --> `🍕`.

This is intended to be a good example for beginners, and can be used as a "make your first add-on" tutorial. As such, the code is well commented (including reference links) and strives to be efficient in its implementation so as to address the following concern stated in the issue:

> we should make sure the replacement script has decent performance, since it could end up being replicated multiple times.

# Concerns / Known Issues / Tradeoffs

  * This example is written in ES6, specifically using `let`, `const`, and `Map`. It could easily be converted to ES5 but since [ES6 support is good across target browsers](http://kangax.github.io/compat-table/es6/), and the stated purpose was to help beginners, I thought we might as well get them using ES6 😄 .
  * For efficiency, this solution employs regular expressions, which can be confusing for a beginner.
    * To alleviate this concern, the regex used is not complicated and I linked to [MDN RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) for reference.
  * The regular expression used matches on substrings. For example, `man` is replaced with `👨`, and while the word `man` is replaced correctly throughout the document, so too are words like `many` (`👨y`), `manipulate` (`👨ipulate`), `command` (`com👨d`), etc.
    * I almost re-wrote this out, but it's actually kind of fun 😄 

Feedback welcome, happy hacktoberfest! 🎃 👻 🎃 👻


emoji-substitution/emojiMap.js
```diff
@@ -0,0 +1,81 @@
+/*
```
aterranova-bv: I split this file out because it could potentially get very lengthy and splitting it out improved readability and maintenance.

emoji-substitution/manifest.json
```diff
@@ -0,0 +1,24 @@
+{
+  "manifest_version": 2,
+  "applications": {
+    "gecko": {
+      "strict_min_version": "49.0.1"
```
aterranova-bv: The `strict_min_version` could probably be lower than this, but I didn't test on anything less than `49.0.1`.

It does have a hard limit at `47`, as that's when [let](http://caniuse.com/#search=let) and [const](http://caniuse.com/#search=const) were added.

examples.json
```diff
@@ -208,6 +208,15 @@
...
   {
+    "name": "emoji-substitution",
+    "description": "Replaces words with emojis.",
+    "url": "https://github.com/mdn/webextensions-examples/tree/master/emoji-substitution",
+    "manifest_keys": [
+      "content_scripts"
+    ],
+    "javascript_modules": []
+  },
+  {
```
aterranova-bv: I wasn't sure if this file was auto-generated or not, but I added Emoji Substitution here for good measure.

emoji-substitution/manifest.json
freaktechnik: It could go way below that, Firefox supports let and const the way the spec suggests since Firefox 36 and has had them since a way longer time (see also https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#Browser_compatibility). I'd assume the hard limit is more about when WebExtensions were introduced...

emoji-substitution/substitute.js
```diff
@@ -0,0 +1,67 @@
...
+    // This node only contains text.
+    // @see https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType.
+
+    // Because DOM manipulation is slow, we don't want to keep setting
+    // textContent after every replacement. Instead, manipulate a copy of
+    // this string outside of the DOM and then perform the manipulation
+    // once, at the end.
+    let content = element.textContent;
+
+    // Replace every occurance of 'word' in 'content' with its emoji.
```
wbamberg: occurance->occurrence

emoji-substitution/manifest.json
wbamberg: 49 is fine I think, since that's the current release.

wbamberg:
This looks really excellent to me :). It seems to work great and the code's clean, well-commented and easy to understand.

@jvillalobos , you mentioned that it needs to have good performance - can you assess that?

emoji-substitution/substitute.js
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @param  {Node} element - The target DOM element.
+ * @return {void}         - Note: the emoji substitution is done inline.
+ */
+function replaceText (element) {
+  // Setting textContent on a node removes all of its children and replaces
+  // them with a single text node. Since we don't want to alter the DOM aside
+  // from substituting text, we only substitute on single text nodes.
+  // @see https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent
+  if (element.nodeType === Node.TEXT_NODE) {
```
aterranova-bv: Ok, will do.

emoji-substitution/substitute.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    // Now that all the replacements are done, perform the DOM manipulation.
+    element.textContent = content;
+  }
+  else {
+    // This node contains more than just text, call replaceText() on each
+    // of its children.
+    for (let i = 0; i < element.childNodes.length; i++) {
+      replaceText(element.childNodes[i]);
```
aterranova-bv: Good points, I like the mutation observer approach.

I think there's a valid trade-off question here for sure. Since this is primarily intended for beginners, making it work on popular sites like Facebook or Twitter seems like a good idea.

On the other hand it does complicate what is intended to be a "make your first add-on" example, and is perhaps a logical "Step 2", or "make your *second* add-on tutorial" 😛 . It could also introduce performance hits on sites that continuously update content - I'm not sure there's a good time to disconnect the observers.

I'll leave it to your team to decide, I'm cool with adding them if y'all want.

aterranova-bv: > Also: the fact that dealing with dynamic content is more complicated seems to make it more important, not less, to have a good-quality example to copy.

Good point.  Ok, I'm on it!

aterranova-bv:
## PR Updated
  * Monitors DOM for additions after initial load
  * Fixes typos

> Monitors DOM for additions after initial load

I used https://twitter.com/hashtag/cats?f=tweets&vertical=default as a testing page. Before this change, new tweets would not have their text replaced.  After this change, they do 🎉 .

I originally also was monitoring for `characterData` changes, but this proved too costly in terms of performance (browser slowed to a crawl) so it is no longer included. This means the algorithm will not run on text nodes if they have their text changed dynamically.

## Not addressed (yet)
> I think there are cases where this replaces text in a textbox, like in Bugzilla or Wordpress. Might be worth checking.

I couldn't get this to happen but if you have a specific page I'll test it out.

@wbamberg  @jvillalobos

emoji-substitution/emojiMap.js
```diff
@@ -0,0 +1,81 @@
...
+dictionary.set('sun', '☀️');
+dictionary.set('sunglasses', '😎');
+dictionary.set('surprised', '😮');
+dictionary.set('tree', '🌲');
+dictionary.set('trophy', '🏆');
+dictionary.set('win', '🏆');
+dictionary.set('wind', '💨');
+dictionary.set('wine', '🍷');
+dictionary.set('wink', '😉');
+dictionary.set('woman', '👩');
```
kumar303: Are these rules ordered? The regex for `woman` needs to be applied before `man`. Currently it looks like the word woman just becomes wo👨 so I guess the `man` regex is getting applied first.

aterranova-bv: Good catch, I totally forgot about that.  I put them in alphabetical order here for readability, and the code simply applies them in index order.

Open to suggestions on improving this.

aterranova-bv: One thing I just thought of is `substitution.js` could sort these by length at the very beginning so even if some words are subsets of others (as is the case with "man" and "woman"), "woman" would be applied first.

This way it wouldn't matter what order they appear in `emojiMap.js`.

emoji-substitution/substitute.js
jvillalobos: I was able to reproduce this with `textarea` elements in WordPress editing tools. If I open a published post that has an Except, the text in the Excerpt field has its words replaced. So if I save it for whatever reason, the emojis would end up being published. So, we need to at least discard that type of element.

aterranova-bv:
PR Updated:

* Fixes substitution ordering bug
* No longer replaces in `textarea` nodes

@jvillalobos , can you verify that this fixes the issue in Wordpress?

jvillalobos:
Yeah, looks good to me. Thanks! r+

wbamberg:
Very nice example :-). Thanks!

emoji-substitution/substitute.js
```diff
@@ -0,0 +1,93 @@
...
+const emojiMap = sortedEmojiMap;
+
+/*
+ * For efficiency, create a word --> search RegEx Map too.
+ */
+let regexs = new Map();
+for (let word of emojiMap.keys()) {
+  // We want a global, case-insensitive replacement.
+  // @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp
+  regexs.set(word, new RegExp(word, 'gi'));
```
wagnerand: I think you'd want `/\bword\b/`

aterranova-bv: Yeah, I went back and forth on this one.  From the PR description:

> The regular expression used matches on substrings. For example, man is replaced with 👨, and while the word man is replaced correctly throughout the document, so too are words like many (👨y), manipulate (👨ipulate), command (com👨d), etc.
I almost re-wrote this out, but it's actually kind of fun 😄

wagnerand: Hm, I think especially for short and common patterns like _man_, this sounds a bit too aggressive to me. It's cool when for things like _mankind_, but words that contain _man_ but have no relation to the concept of a _man_ could cause some confusion.

atsay:
Thanks so much for this example, Andrew! We've added your contribution [here](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#October_2016).

aterranova-bv:
> Thanks so much for this example, Andrew! We've added your contribution here.

That's awesome, thanks!

emoji-substitution/substitute.js
wbamberg: FWIW, I think the current behavior here is fine. I read your description in the PR when you first made it, and agree that it's kind of fun the way it is now. It makes the add-on a kind of rebus-generator :).

kumar303:
One thing that would be nice to do in a follow-up patch is to use an array for the pattern sorting instead of a new map. The thing with maps is that the order of its keys is [not predictable](http://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order). On some random page load, the order of the keys may be different.

PR #112, opened 2016-10-16
Use browser_style in webpack example

freaktechnik:
This uses the from browser_styles in the popup. I had to add some CSS to make it look good, since the browser styles don't cover everything.


wbamberg:
@bwinton , do you have any feedback on this? It looks good to me.

bwinton:
Code seems good to me.  I'ld love it if @freaktechnik could dig into it a little more and tell me why  the footer creates a horizontal overflow so that we can possibly fix it in the styles, but that shouldn't block landing this.  :)

wbamberg:
Thanks!

PR #113, opened 2016-10-18
Added zoom api example to tabs-tabs-tabs

DarKFlameS:
Added some zoom manipulation example in the tabs-tabs-tabs as suggested in the Contribution guide. Made as a participation in "Hacktoberfest"

### DIFF
--- a/tabs-tabs-tabs/tabs.html
+++ b/tabs-tabs-tabs/tabs.html
@@ -7,16 +7,34 @@
   </head>
 
 <body>
-  <a href="#" id="tabs-move-beginning">Move active tab to the beginning of the window</a><br>
-  <a href="#" id="tabs-move-end">Move active tab to the end of the window</a><br>
-  <a href="#" id="tabs-duplicate">Duplicate active tab</a><br>
-  <a href="#" id="tabs-reload">Reload active tab</a><br>
-  <a href="#" id="tabs-remove">Remove active tab</a><br>
-  <a href="#" id="tabs-create">Create a tab</a><br>
-  <a href="#" id="tabs-alertinfo">Alert active tab info</a><br>
-  <a href="#" id="tabs-add-zoom">Zoom in</a><br>
-  <a href="#" id="tabs-decrease-zoom">Zoom out</a><br>
-  <a href="#" id="tabs-default-zoom">Reset zoom</a><br>
+  <div class="panel">
+    <div class="panel-section panel-section-header">
+      <div class="text-section-header">Tabs-tabs-tabs</div>
+    </div>
+
+    <a href="#" id="tabs-move-beginning">Move active tab to the beginning of the window</a><br>
+    <a href="#" id="tabs-move-end">Move active tab to the end of the window</a><br>
+
+    <div class="panel-section-separator"></div>
+
+    <a href="#" id="tabs-duplicate">Duplicate active tab</a><br>
+    <a href="#" id="tabs-reload">Reload active tab</a><br>
+    <a href="#" id="tabs-alertinfo">Alert active tab info</a><br>
+
+    <div class="panel-section-separator"></div>
+
+    <a href="#" id="tabs-create">Create a tab</a><br>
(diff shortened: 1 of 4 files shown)


tabs-tabs-tabs/tabs.html
```diff
@@ -14,6 +14,9 @@
   <a href="#" id="tabs-remove">Remove active tab</a><br>
   <a href="#" id="tabs-create">Create a tab</a><br>
   <a href="#" id="tabs-alertinfo">Alert active tab info</a><br>
+  <a href="#" id="tabs-add-zoom">Increase tab zoom</a><br>
+  <a href="#" id="tabs-decrease-zoom">decrease tab zoom</a><br>
+  <a href="#" id="tabs-default-zoom">default tab zoom</a><br>
```
wbamberg: "default tab zoom" -> "Reset zoom"

tabs-tabs-tabs/tabs.js
```diff
@@ -70,5 +70,49 @@ document.addEventListener("click", function(e) {
     });
   }
 
+  else if (e.target.id === "tabs-add-zoom") {
+      callOnActiveTab((tab) => {
```
wbamberg: The rest of this file uses 2-space tabs, could you adjust the new code to be consistent with that?

tabs-tabs-tabs/tabs.js
```diff
@@ -70,5 +70,49 @@ document.addEventListener("click", function(e) {
...
+        chrome.tabs.getZoom(tab.id, function(zoomFactor){
+            var newZoomFactor = zoomFactor + 0.2;
```
wbamberg: It would be nice to define constants for things like this.

e.g. 
const ZOOM_INCREMENT = 0.2;
const MAX_ZOOM = 3;
const MIN_ZOOM = 0.3;
const DEFAULT_ZOOM = 1;

tabs-tabs-tabs/tabs.js
```diff
@@ -70,5 +70,49 @@ document.addEventListener("click", function(e) {
...
+            }
+        });
+      });
+  }
+
+  else if (e.target.id === "tabs-decrease-zoom") {
+      callOnActiveTab((tab) => {
+        chrome.tabs.getZoom(tab.id, function(zoomFactor){
+            var newZoomFactor = zoomFactor - 0.2;
+            newZoomFactor = newZoomFactor < 0.3 ? 0.3 : newZoomFactor
```
wbamberg: You could move these 2 lines inside the "else" at line 98, since you don't need to do this if you are already at max.

wbamberg:
Thanks for the contribution. It's great to have some extra tab methods demonstrated here. I just had a few minor comments.

One other thing - do you think it would be possible to get a little padding/margin around the links? It looks quite cramped in this popup.

If you really wanted to get fancy, you could experiment with using the "browser_style" property in "browser_action": https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/browser_action. This will include a built-in stylesheet (this one: chrome://browser/content/extension.css), and you can use that to get fancy styles for popups. There's some documentation of the styles you can get here: https://firefoxux.github.io/StyleGuide/#/controls
There's an example add-on that uses it here:
https://github.com/mdn/webextensions-examples/tree/master/latest-download

But this part is strictly optional :).

DarKFlameS:
Ok, I'll work on these changes soon!

DarKFlameS:
I've made the fancy part, so the extension panel now looks better. I've change the zoom links to buttons, so it would have the buttons as an example of the style guide too.

wbamberg:
I think this looks so much better. I especially like the way you've grouped the different operations. But, I think it's weird to make the zoom operations buttons, and the others links. I think they should all be the same kind of control (now I look at it, I think they should just be links).

DarKFlameS:
Ok, I'll turn it all back to links

wbamberg:
Thanks @DarKFlameS !

Issue #115, opened 2016-10-19
Get rid of innerHTML


wagnerand:
Reviewers discourage the use of `innerHTML` for security and performance reasons. Let's live up to our own best practices. :)

There are currently only two occurrences:
```
history-deleter/history.js
17:  document.getElementById('history').innerHTML = `No history for ${hostname}.`;

navigation-stats/popup.js
17:  listEl.innerHTML = "";
```

Hrily:
I would like to work upon it...
Is it okay to use DOM methods?

wbamberg:
> I would like to work upon it...

Great!

> Is it okay to use DOM methods?

Yes!

Hrily:
Can you tell me what does the ${} operator do?
I'm not able to get it.

wbamberg:
It's for interpolation in template literals: "Template literals are string literals allowing embedded expressions."

-> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals.

Hrily:
So, \`\` no different than a "" string?

wbamberg:
No, it is different.

`` is used to create a template literal, which enables you to do string interpolation:

```
var sub = "World";
var greet1 = `Hello ${sub}` // -> "Hello World"
var greet2 = "Hello ${sub}" // -> "Hello ${sub}"
```

Hrily:
Okay, I didn't know about them...

Hrily:
I have made a PR. #116 
Can you review it?

wbamberg:
Merged https://github.com/mdn/webextensions-examples/pull/116. Thanks @Hrily !

Hrily:
Welcome :) 
Even if it is a small modification, I'm feeling proud as my first Open Source contribution is to the MDN :D

PR #116, opened 2016-10-20
Removed innerHTML

Hrily:
Get rid of innerHTML #115 
Done.. :+1:

### DIFF
--- a/history-deleter/history.js
+++ b/history-deleter/history.js
@@ -17,7 +17,7 @@ function no_history(hostname) {
   var history_text = document.getElementById('history');
   while(history_text.firstChild )
 	history_text.removeChild(history_text.firstChild);
-  history_text.appendChild(document.createTextNode(`No history for ${hostname}.`) );
+  history_text.appendChild(document.createTextNode("No history for " + hostname + ".") );
 }
 
 // When the page is loaded find the current tab and then use that to query
--- a/history-deleter/history.js
+++ b/history-deleter/history.js
@@ -15,6 +15,8 @@ function set_domain(domain) {
 
 function no_history(hostname) {
   var history_text = document.getElementById('history');
+  while(history_text.firstChild)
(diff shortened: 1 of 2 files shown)


history-deleter/history.js
```diff
@@ -14,7 +14,10 @@ function set_domain(domain) {
 }
 
 function no_history(hostname) {
-  document.getElementById('history').innerHTML = `No history for ${hostname}.`;
+  var history_text = document.getElementById('history');
+  while(history_text.firstChild )
+	history_text.removeChild(history_text.firstChild);
```
wagnerand: Is this necessary? Does the `history` element ever contain child nodes?

history-deleter/history.js
```diff
@@ -14,7 +14,10 @@ function set_domain(domain) {
...
+  history_text.appendChild(document.createTextNode("No history for " + hostname + ".") );
```
wagnerand: * It would be cool if you kept using template strings.
* Also, why not just use `.textContent`?
* Did you check `hostname` is just plain-text?

history-deleter/history.js
Hrily: I didn't see the history element... :D

history-deleter/history.js
Hrily: Is it now fine?...

history-deleter/history.js
Hrily: Okay...
Will do it...

Hrily: Done...

wbamberg:
@Hrily , this looks fine to me, but I'd like to get @wagnerand 's word too, since he filed the original bug. Thanks for your work on it!

Hrily:
Your Welcome...  :)

wbamberg:
Thanks @Hrily ! :-)

PR #118, opened 2016-10-24
Added chrome.tabs.onRemoved sample

DarKFlameS:
Shows how to use the "chrome.tabs.onRemoved" and the removeInfo object


wbamberg:
I very much like the idea of demoing the tabs event handler in this example, but I'm not so keen on making them separate commands in the UI.

Why not just add the event listeners in the popup script, and have them be triggered naturally when the user selects an action? For example:

```
chrome.tabs.onRemoved.addListener(function(tabId, removeInfo){
  console.log(`Tab ${tabId} is closing`);
  if(removeInfo.isWindowClosing) {
    console.log(`Its window is also closing`);
  } else {
    console.log(`Its window is not closing`);
  }
});
```

...then it would just trigger when the user selects the existing "Remove active tab" option. We could have similar listeners for the "Move active tab..." and so on.

Also I think using `alert()` is a bit aggressive - `console.log()` works better.

What do you think?

DarKFlameS:
I've made the changes, and added the onMoved event listener too. I've just used alert to make thinks more "visible", but I've now changed it all to console.log.

tabs-tabs-tabs/tabs.js
```diff
@@ -124,3 +124,21 @@ document.addEventListener("click", function(e) {
...
 });
+
+//onRemoved listener. fired when tab is removed
+chrome.tabs.onRemoved.addListener(function(tabId, removeInfo){
+  console.log(`The tab with id: ${tabId}, is closing`);
+
+  if(removeInfo.isWindowClosing) {
+    console.log(`Its window is also closing.`);
+  } else {
+    console.log(`Its window window is not closing`);
```
wbamberg: "window window" -> "window" :)

tabs-tabs-tabs/tabs.js
```diff
@@ -124,3 +124,21 @@ document.addEventListener("click", function(e) {
...
+  }
+});
+
+//onMoved listener. fired when tab is moved into the same window
+chrome.tabs.onMoved.addListener(function(tabId, moveInfo){
+  var startIndex = moveInfo.fromIndex;
+  var endIndex = moveInfo.toIndex;
+  console.log(`Tab with id: ${tabId} moved from index: ${startIndex} to: ${endIndex}`);
```
wbamberg: Super picky, but since you'll need to fix the typo above anyway...

...this gives output like:

`Tab with id: 4 moved from index: 2 to: 0`

I thing it would be better to repeat "index", like:

`Tab with id: 4 moved from index: 2 to index: 0`

wbamberg:
Thanks again @DarKFlameS !

Issue #71, opened 2016-06-22
Remove id from the examples


andymckay:
Given that we expect the examples to be copied and re-used, we should probably remove the applications section, or at least the id from all the add-ons. That way developers can copy them and get them signed on AMO and check out the whole process without having to get into ID collisions.

Maybe wait till web-ext lands with no id support: mozilla/web-ext#178

wbamberg:
@andymckay - do you think we should keep the strict_min_version (and by implication the applications key)?

(I think probably we should, because people get into enough difficulties with versions as it is. The downside of course is that it provokes warnings in Chrome.)

andymckay:
Good point, we should. Agreed.

kumar303:
If you leave in the strict_min_version, beware the AMO will refuse to sign it without an ID when you require a Firefox less than 47:

    Server response: GUID is required for Firefox 47 and below.

wbamberg:
@kumar303 : is that <47, or <=47?

Your comment suggests <, but your quote suggests <=.

If 47 is OK, then I'd probably suggest updating strict_min_version to 47 for every example (except those that need 48 or 49 of course). But if 47 is not OK, it's a bit tricker, because 47 is the current release Firefox.

kumar303:
Hmm. Yes, the error message makes it seem like <= 47 but I haven't tried uploading to AMO to verify.

kumar303:
Scratch that. I tested it on AMO. To upload an extension without an ID *and* a `strict_min_version` you have to set `strict_min_version = 48`

wbamberg:
To summarise, options are:

**1) remove "applications" entirely.**

Pros:
* add-ons will get automatic IDs, so no ID collisions if people copy these add-ons and submit them to AMO.
* add-ons will install in Chrome with no warnings.

Cons:
* there will be no "strict_min_version", so people won't get a clue when they install, say, favourite-colour on Firefox 47. We could document version requirements more obviously, to try to help with that.

**2) remove "applications/id", but keep "applications/strict_min_version" the same**

Pros:
* add-ons will get automatic IDs, so no ID collisions if people copy these add-ons and submit them to AMO.
* users will get a clue about version compatibility (via "applications/strict_min_version")

Cons:
* if "applications/strict_min_version" is <48, then AMO will give an error.

**3) remove "applications/id", keep "applications/strict_min_version", and set it to Firefox 48 or higher**

Pros:
[...]

PR #123, opened 2016-10-27
Webextension example of some windows api functions.

DarKFlameS:


window-manipulate/window.js
```diff
@@ -0,0 +1,60 @@
...
+      chrome.windows.update(currentWindow.id, updateInfo);
+    });
+  }
+
+  else if (e.target.id === "window-create-incognito") {
+    createData = {
+      incognito: true,
+    };
+    chrome.windows.create(createData, function(){
+      alert("The incognito window has been created");
```
wbamberg: Is there a reason you've included this alert()?

window-manipulate/manifest.json
```diff
@@ -0,0 +1,26 @@
+{
+  "applications": {
+    "gecko": {
+      "id": "window-manipulate@mozilla.org",
+      "strict_min_version": "47.0a1"
+    }
+  },
+  "browser_action": {
+    "browser_style": true,
+    "default_title": "Window manipulate",
```
wbamberg: Could we use "Window manipulator" or something like that instead? Here, and also in the title of the popup, and the "name" field.

window-resize/manifest.json
```diff
@@ -0,0 +1,26 @@
+{
```
wbamberg: I think you didn't mean to include this file in the PR, right?

wbamberg:
Looks great!

I had a few minor comments. Also, the icons:

- it would be nice if they all have a common naming scheme, instead of two being "default*" and two being "window*". How about "window19" etc?

- because the icons are black, they are invisible in the dark Firefox theme, which is the default theme in Firefox Developer Edition (if you have Dev Edition and to have switched the theme back to normal see the dark theme by visiting "about:addons", switching to "Appearance", and enabling the "Developer Edition" theme).

window-manipulate/manifest.json
```diff
@@ -0,0 +1,26 @@
+{
+  "applications": {
```
kumar303: In https://github.com/mdn/webextensions-examples/issues/71 we decided to not specify IDs for new extensions. This will help preserve compatibility with Chrome. You can remove this entire `applications` key from the manifest.

window-manipulate/window.js
```diff
@@ -0,0 +1,60 @@
...
+    chrome.windows.getCurrent(
+      function(currentWindow) {
+        callback(currentWindow);
+      }
+    );
+  }
+
+  if (e.target.id === "window-update-size_768") {
+    callOnCurrentWindow((currentWindow) => {
+      updateInfo = {
```
kumar303: This and all other declarations should be scoped locally to prevent any unintended side effects. For example:
````
var updateInfo = {
}
````

File: window-manipulate/window.js [hunk 6]
```diff
@@ -0,0 +1,60 @@
+document.addEventListener("click", function(e) {
+
+  function callOnCurrentWindow(callback){
+    chrome.windows.getCurrent(
+      function(currentWindow) {
```
kumar303: Since you're already using arrow functions elsewhere, I'd suggest using them everywhere because they make the code more concise.

wbamberg:
Thanks again @DarKFlameS ! Nice work :).

PR #124, opened 2016-10-29
Selection to clipboard

iampeterbanjo:
An example using the [clipboard API](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Interact_with_the_clipboard)


selection-to-clipboard/README.md
```diff
@@ -0,0 +1,14 @@
+# selection-to-clipboard
+
+## What it does
+
+This extension includes:
+
+* a content script, "content-script.js", that is injected into all pages
+
+The content script listens for text selections in the page it's attached to and copies the text to the clipboard.
```
andymckay: This doesn't explain that it only does this on mouse up, might be worth commenting on that.

selection-to-clipboard/manifest.json
```diff
@@ -0,0 +1,25 @@
...
+    "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/selection-to-clipboard",
+
+    "icons": {
+        "48": "icons/clipboard-48.png"
+    },
+
+    "applications": {
+        "gecko": {
+            "id": "selection-to-clipboard@mozilla.org",
+            "strict_min_version": "49.0a1"
```
andymckay: I think this was fixed in https://bugzilla.mozilla.org/show_bug.cgi?id=1197451, which landed in Firefox 51. I tried this in nightly and it worked, but it didn't in release, Firefox 49. So I think we should increase this strict_min_version to a working Firefox version.

selection-to-clipboard/manifest.json
```diff
@@ -0,0 +1,25 @@
...
+    "version": "1.0",
+    "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/selection-to-clipboard",
+
+    "icons": {
+        "48": "icons/clipboard-48.png"
+    },
+
+    "applications": {
+        "gecko": {
+            "id": "selection-to-clipboard@mozilla.org",
```
andymckay: I think we've been trying to not specify an id in the add-ons, they can be run locally in Firefox and if someone wants to sign them, they can do so without conflicting with anyone else.

selection-to-clipboard/manifest.json
```diff
@@ -0,0 +1,25 @@
+{
+    "manifest_version": 2,
+    "name": "selection-to-clipboard",
+    "description": "Example of WebExtensionAPI for writing to the [clipboard](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Interact_with_the_clipboard) ",
```
andymckay: Putting markdown in the description isn't supported by Firefox, it looks like this. Could you move that to readme please? 

screenshot 2016-10-31 11 33 45

selection-to-clipboard/manifest.json
iampeterbanjo: Thanks. Bumped to 51

selection-to-clipboard/manifest.json
iampeterbanjo: Ok. Removed and thanks for the tip

selection-to-clipboard/manifest.json
iampeterbanjo: Moved to README.md

selection-to-clipboard/README.md
iampeterbanjo: I've updated the text

andymckay:
I think this looks good, @wbamberg any thoughts?

wbamberg:
I think this is great. I'm really pleased to have an example of clipboard interaction. Thanks!

However, "51" won't work as strict_min_version, because "clipboardWrite" is currently in Firefox Dev Edition, whose version number is "51.0a2" or something like that. "51" is considered a higher version than "51.0a2", because "51.0a2" is a pre-release.

So this won't install in Dev Edition. You need strict_min_version to be "51.0a1".

wbamberg:
Also, this seems to be an example of using the "clipboardWrite" permission, but I think you don't actually need that permission in this case, because the copy is done inside an event handler for a user action. Try removing the permission, the example will still work.

Conversely, try setting a timeout in the event handler, and doing the copy in the timeout:

```
document.addEventListener("mouseup", () => {
  setTimeout(copySelection, 1);
});
```

Now you will need "clipboardWrite". This should be covered in https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Interact_with_the_clipboard.

wbamberg:
I was thinking a bit about this, what would be a nice example of using copy outside an event handler. Maybe an add-on could be a page action that's shown only in Wikipedia pages, that, when clicked, injects a content script that copies the first paragraph of the article?

I think something like this `document.querySelector("#mw-content-text>p").textContent` would get the first paragraph, and this trick: https://stackoverflow.com/questions/3436102/copy-to-clipboard-in-chrome-extension/12693636#12693636 works to copy content without needing it to be selected.

Just a suggestion, I'm sure there are lots of other possibilities. But I do think it's worth having an example that requires "clipboardWrite", or adapting this one so it does.

iampeterbanjo:
Thanks @wbamberg. I've bumped the versions and wrapped the function in a timeout to force the permissions. It works as you described.

I like your suggestion about the Wikipedia app. One app I have been investigating how to build is passing page text into a speech synthesis engine to read it out load. ChromeVox and SpeakIt extensions for Google chrome force the user to select the text which is limiting for long pages especially above/below the fold. I still don't know how to create a HTML node selector like the browser element inspector to visually select paragraphs of text - that would be nice :)

wbamberg:
Thanks for making the updates!

Sorry if I wasn't clear: I hadn't really intended the timeout suggestion as a suggested way to write this example, just as a way to show you when the permission is needed.

I think it looks strange to have this in the example, it would make me think, why are we doing the copy in a timeout? It seems like an artificial way to implement this add-on.

My Wikpedia suggestion was an attempt to come up with a problem whose solution would _naturally_ require "clipboardWrite". I don't think this example has to be the Wikipedia one. It could be, or it could be a different add-on with a similar structure (for example, any add-on where (1) the user clicks a browser/page action, (2) that injects a script, (3) that copies some text), or it could be a totally different add-on, that also naturally requires clipboardWrite.

Does that sound reasonable?

```
 One app I have been investigating how to build is passing page text into a speech synthesis engine to read it out load. ChromeVox and SpeakIt extensions for Google chrome force the user to select the text which is limiting for long pages especially above/below the fold. I still don't know how to create a HTML node selector like the browser element inspector to visually select paragraphs of text - that would be nice :)
```

That's a really interesting idea! I guess you could have a content script that listens for mouseover events and highlights the element you are over, then listens for clicks and sends the text back to the add-on to send to speech synthesis.

It sounds like it might be out of scope for this example though :).

iampeterbanjo:
No worries - that was my fault because @wbamberg you didn't make a change request.

Should I remove the clipboard permissions and the timeout function?

wbamberg:
> Should I remove the clipboard permissions and the timeout function?

I think that's the simplest option. I think if you do that it would be worth adding a comment that in this case you don't need the permission because it's a response to a user action in the page, and linking to https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Interact_with_the_clipboard.

The alternative would be to build a different add-on that does require the permission, as we've discussed.

I think either would be useful :).

iampeterbanjo:
@wbamberg thanks - I've made that update

wbamberg:
Thanks @iampeterbanjo !

iampeterbanjo:
@wbamberg thank you! :+1:

PR #125, opened 2016-10-29
create example for switching tabs

iampeterbanjo:
Created an example for switching tabs in the current window


andymckay:
Demonstrating how to use active is cool and I think that's a good addition. However I think in this case, it probably makes more sense to add to the tabs-tabs-tabs addon as an additional action.

iampeterbanjo:
Ok. Thanks for the suggestion

iampeterbanjo:
I've created another PR so closing this one :)

PR #126, opened 2016-10-29
Added example for the highlight method.

rahulyesantharao:
Highlights the current tab and the next tab.

### DIFF
--- a/tabs-tabs-tabs/tabs.js
+++ b/tabs-tabs-tabs/tabs.js
@@ -121,8 +121,8 @@ document.addEventListener("click", function(e) {
       });
     });
   }
-
-  else if (e.target.id === "tabs-highlight") { // highlights current tab and next tab
+  // Currently (11/2/2016) only supported by Chrome
+  else if (e.target.id === "tabs-highlight") { // highlights current tab and next tab (cycles back to first tab if current tab is the last one)
     callOnActiveTab((tab, tabs) => {
       next = (tab.index+1) % tabs.length;
       chrome.tabs.highlight({tabs:[tab.index, next]});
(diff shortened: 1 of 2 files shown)


tabs-tabs-tabs/tabs.js
```diff
@@ -122,5 +122,12 @@ document.addEventListener("click", function(e) {
     });
   }
 
+  else if (e.target.id === "tabs-highlight") { // highlights current tab and next tab
```
andymckay: This doesn't actually work on anything but Chrome, so I think we should have some note about that somewhere, maybe in the readme or in the panel itself. 

https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/highlight

tabs-tabs-tabs/tabs.js
```diff
@@ -122,5 +122,12 @@ document.addEventListener("click", function(e) {
...
+    callOnActiveTab((tab, tabs) => {
+      next = (tab.index+1) % tabs.length;
```
andymckay: This isn't the next tab. If tab.index + 1 is equal to tabs.length, its 0, which is likely the first tab. I can see the need to pick another tab though, but the comment on the line above is inaccurate and could maybe be improved?

tabs-tabs-tabs/tabs.js
rahulyesantharao: Oh yeah I noticed that, but wasn't sure what to do about it; I'll add the comment in!

rahulyesantharao:
Hi! I think I made the changes, but I'm not sure if I did it right because I'm new to GitHub.  Please let me know if anything didn't work.  Thanks!

andymckay:
Looks good to me, @wbamberg ?

wbamberg:
Looks great, thanks @rahulyesantharao !

Issue #117, opened 2016-10-21
Create example of using eslint to ensure Firefox compatibility


kumar303:
Firefox supports most of ES6 but not all of it. For example, it doesn't support the [import statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) at the time of this writing.

Let's create an example of an extension that is configured with eslint to make sure that its code uses only JavaScript that is supported by Firefox. This example extension would not [use a transpiler](https://github.com/mdn/webextensions-examples/issues/97) on the code.

I think there are some pre-configured lint rules to use, such as https://www.npmjs.com/package/eslint-plugin-mozilla

kumar303:
Thanks again for this @iampeterbanjo ! This landed in https://github.com/mdn/webextensions-examples/pull/128

PR #128, opened 2016-10-30
Eslint example

iampeterbanjo:
Addresses #117 but I’m not sure if the eslint configuration is correct because the package eslint-plugin-mozilla isn’t documented. I used other eslint docs and plugins as a guide.

Should this example have commented versions of code that work and don’t work to show eslint is setup properly? I used an example with es6 `import` as a start.


kumar303:
Hi, thanks for kicking this of! This is a great start. Here are a few other things I think it will need:

* This should be a functional WebExtension that you can install. You could maybe [add a button](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Add_a_button_to_the_toolbar) that calls `getUsefulContents()` to show some data but any minimal example would do.
  * This will allow you to test out the features that won't work in Firefox. For example, I think the `import` statement won't work in Firefox.
* The README should show you how to execute the lint command. I think if you add an [npm script](https://docs.npmjs.com/misc/scripts) that aliases `eslint` it could be as simple as `npm run lint`.

iampeterbanjo:
I've added a npm script to run a local eslint config and made it into a simple browserAction extension. I've left some comments in to show import/export can be used instead which i used to test that the linter was working.

eslint-example/README.md
```diff
@@ -0,0 +1,7 @@
+# ESLint Example
+
+## What it shows
+
+A project configured to use the [Mozilla ESlint configuration](https://www.npmjs.com/package/eslint-plugin-mozilla)
+
+* to run locally configured eslint use `npm run lint`
```
kumar303: This is missing the instruction `npm install`

eslint-example/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "description": "",
+    "main": "index.js",
+    "dependencies": {},
+    "devDependencies": {
+        "eslint": "^3.9.0",
+        "eslint-plugin-mozilla": "^0.2.3"
+    },
+    "scripts": {
+        "test": "echo \"Error: no test specified\" && exit 1",
+        "eslint": "./node_modules/.bin/eslint",
```
kumar303: This isn't necessary. The `npm run lint` command will already work fine as you have it because npm magically puts `./node_modules/.bin` on PATH for scripts.

eslint-example/README.md
```diff
@@ -0,0 +1,7 @@
+# ESLint Example
+
+## What it shows
+
+A project configured to use the [Mozilla ESlint configuration](https://www.npmjs.com/package/eslint-plugin-mozilla)
```
kumar303: I think another sentence will be helpful: "By using a linter on your extension source code you can be sure that you will only be writing JavaScript that is compatible with Firefox."

eslint-example/main.js
```diff
@@ -0,0 +1,12 @@
+ // not supported in Firefox
+ // import { getUsefulContents } from "file";
```
kumar303: Interesting. In Firefox I see an error that suggests this *might* work. I don't think it will work yet though. I also tried it in Nightly, same error.

eslint-example/manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+    "manifest_version": 2,
+    "description": "Example using eslint",
+    "name": "eslint-example",
+    "version": "1.0",
+    "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/eslint-example",
+
+    "applications": {
```
kumar303: This applications section can be removed. It's not needed for this example so removing it would be more helpful to someone copying/pasting this code. When you omit this section, the extension can be compatible with Chrome.

kumar303:
@Standard8 can you suggest any other changes that would make this is a useful linting example? The goal is to show how one can use eslint to ensure that their code is compatible with Firefox (without a transpiler).

Standard8:
> @Standard8 can you suggest any other changes that would make this is a useful linting example? The goal is to show how one can use eslint to ensure that their code is compatible with Firefox (without a transpiler).

For the [example webextension repository](https://github.com/Standard8/example-webextension) I've been working on, I used a separate sub-directory for the add-on to separate it from tools/tests and this is combined with the use of web-ext.

So in that repository, `npm run lint` actually runs the web-ext as well as eslint.

I also have a bunch of useful rules enabled by default, as a starting point for new folks.

eslint-example/package.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "eslint-example",
+    "version": "1.0.0",
+    "description": "",
+    "main": "index.js",
+    "dependencies": {},
+    "devDependencies": {
+        "eslint": "^3.9.0",
+        "eslint-plugin-mozilla": "^0.2.3"
```
Standard8: You shouldn't need `eslint-plugin-mozilla` - that's more for the old style bootstrap extensions that access the chrome privileged code.

My example repo has a [list of plugins](https://github.com/Standard8/example-webextension/blob/master/docs/Linting.md) that might be useful.

iampeterbanjo: Thanks. I've removed the mozilla plugin and added `"globals": {"chrome": true}` and `env: { "browser": true }`

eslint-example/manifest.json
iampeterbanjo: Removed

eslint-example/main.js
iampeterbanjo: Ok. I've left it commented out for now.

eslint-example/package.json
iampeterbanjo: I've removed it

eslint-example/README.md
iampeterbanjo: Added a line for `npm install`

.gitignore
```diff
@@ -0,0 +1 @@
+node_modules/
```
kumar303: I just realized that this applies to the entire repo of examples. I don't think it's such a good idea to do that. Can you move it to `eslint-example/.gitignore` ? That will serve the same purpose. That's also a nice place since someone may start by copying the whole folder.

eslint-example/.eslintrc.json
```diff
@@ -0,0 +1,10 @@
+{
+		"globals": {"chrome": true},
```
kumar303: nit: this indentation is inconsistent with the rest of the file.

eslint-example/.eslintrc.json
```diff
@@ -0,0 +1,10 @@
...
+    "rules": {
+        "no-set-state": "off"
+    },
+    "env": {
+    		"browser": true,
```
kumar303: nit: this is also an inconsistent indentation

eslint-example/.eslintrc.json
kumar303: can you also add `"browser": true` ? That's the preferred global to use but `chrome` should be supported by the linter too.

eslint-example/.eslintrc.json
```diff
@@ -0,0 +1,10 @@
...
+        "es6": true
+    }
```
kumar303: I think it's worth add [extends: "eslint:recommended"](https://github.com/Standard8/example-webextension/blob/master/.eslintrc.js#L19) because that catches a lot of standard JS errors.

eslint-example/.eslintrc.json
```diff
@@ -0,0 +1,10 @@
+{
+		"globals": {"chrome": true},
+    "rules": {
+        "no-set-state": "off"
+    },
+    "env": {
+    		"browser": true,
+        "es6": true
```
kumar303: Ah, yes! From the documentation, this appears to be Firefox compatible (everything but imports).

.gitignore
iampeterbanjo: sure. done

eslint-example/.eslintrc.json
iampeterbanjo: no problem. my bad for mixing tabs and spaces

iampeterbanjo: ok

eslint-example/.eslintrc.json
iampeterbanjo: ok

eslint-example/.eslintrc.json
iampeterbanjo: ok - no change here

eslint-example/.eslintrc.json
iampeterbanjo: ok - I've added it

iampeterbanjo:
I've made the requested changes. I also added `/* exported getUsefulContents */` and `/* global getUsefulContents */` to file.js and main.js to pass eslint in a typical scenario of adding multiple script tags to the page with global variables. If there's a better approach I can make changes.

eslint-example/README.md
```diff
@@ -0,0 +1,10 @@
+# ESLint Example
+
+## What it shows
+
+A project configured to use the [Mozilla ESlint configuration](https://www.npmjs.com/package/eslint-plugin-mozilla). By using a linter on your extension source code you can be sure that you will only be writing JavaScript that is compatible with Firefox.
```
kumar303: oops, this opening sentence is no longer true so it will be confusing. I would suggest replacing it with this:

"This is an example of a WebExtension project configured for development with [eslint](http://eslint.org/)."

eslint-example/main.js
```diff
@@ -0,0 +1,13 @@
+/* global getUsefulContents */
```
kumar303: Yes! Using eslint compatible comments is a nice approach

eslint-example/main.js
```diff
@@ -0,0 +1,13 @@
...
+// not supported in Firefox
+// import { getUsefulContents } from "file";
```
kumar303: I don't think it's so helpful to put this comment in here. The linter will complain if someone accidentally adds an import statement. The `es6: true` flag implies that `import` is not supported.

iampeterbanjo:
Thanks

Issue #119, opened 2016-10-25
Example using indexeddb


andymckay:
Maybe an example using indexeddb would be nice? Could maybe integrate with an API we currently don't have an example with. Doesn't need to be complicated, just shows that you can use it in a background page.

https://developer.mozilla.org/en/docs/Web/API/IndexedDB_API

Pheil:
When testing this code with FF 52.0a2 (2017-01-20) I get the following error.  Is access to IndexedDB supported yet?

"IndexedDB request denied"

wbamberg:
@Pheil , I think IndexedDB access from WebExtensions should work just fine in the current Firefox. I haven't tried the add-on associated with this particular issue, but I did try this one: https://github.com/mdn/webextensions-examples/pull/171 (also not merged yet) and it worked fine for me.

biphobe:
@wbamberg Didn't this issue become obsolete after merging #224?

wbamberg:
Yes, looks like it :). Thanks!

PR #129, opened 2016-10-30
Notes with indexeddb

iampeterbanjo:
Create example app to show using indexeddb in background page addresses #119


notes-with-indexeddb/notes.html
```diff
@@ -0,0 +1,36 @@
...
+  <head>
+    <meta charset="utf-8">
+    <title>Notes</title>
+    <link rel="stylesheet" type="text/css" href="notes.css">
+  </head>
+
+<body>
+  <h1>Notes</h1>
+	
+	<form id="add-new-item">
```
kmaglione: Please expand tabs to spaces throughout this file, and use consistent 2-space indentation.

notes-with-indexeddb/notes.html
```diff
@@ -0,0 +1,36 @@
...
+				<input type="text" name="new-description" id="new-description">
+			</div>
+		
+			<button type="submit">Add</button>
+		</fieldset>
+	</form>
+
+  <div id="notes"></div>
+
+  <script src="notes.js"></script>
```
kmaglione: Please move this to the `<head>`

notes-with-indexeddb/notes.html
```diff
@@ -0,0 +1,36 @@
...
+</body>
+
```
kmaglione: Nit: Please remove empty line.

notes-with-indexeddb/manifest.json
```diff
@@ -0,0 +1,27 @@
...
+    },
+
+    "background": {
+        "scripts": ["background.js"]
+    },
+
+    "browser_action": {
+        "default_icon": "icons/notebook-32.png"
+    }
+
```
kmaglione: Nit: Please remove empty line.

notes-with-indexeddb/manifest.json
```diff
@@ -0,0 +1,27 @@
+{
+
```
kmaglione: Nit: Please remove empty line.

notes-with-indexeddb/manifest.json
```diff
@@ -0,0 +1,27 @@
...
+    "description": "Examplifies working with IndexedDB in a background page",
+    "manifest_version": 2,
+    "name": "notes-with-indexeddb",
```
kmaglione: This is meant to be a human-readable name, so it should be something like `"Notes with IndexedDB"`

notes-with-indexeddb/manifest.json
```diff
@@ -0,0 +1,27 @@
...
+    "version": "1.0",
+    "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/notes-with-indexeddb",
+    "icons": {
+        "48": "icons/notebook-48.png"
```
kmaglione: It would be best to include the icons in SVG format so that they scale better for high resolution displays, and when placed in the "hamburger" menu.

notes-with-indexeddb/notes.js
```diff
@@ -0,0 +1,169 @@
+/*
+Feature detection
+ */
+// DON'T use 'var indexedDB = ...' if you're not in a function.
+window.indexedDB = window.indexedDB || window.mozIndexedDB || window.webkitIndexedDB || window.msIndexedDB;
```
kmaglione: No need for this. The bare `window.indexedDB` is supported anywhere this code is capable of running. Same for the other keys.

notes-with-indexeddb/notes.js
```diff
@@ -0,0 +1,169 @@
+/*
```
kmaglione: Please add `"use strict";`

notes-with-indexeddb/background.js
```diff
@@ -0,0 +1,14 @@
+/*
```
kmaglione: Please add `"use strict";` to the top of all scripts.

notes-with-indexeddb/background.js
```diff
@@ -0,0 +1,14 @@
...
+Open a new tab, and load "notes.html" into it.
```
kmaglione: Nit: Please add a ` * ` to the start of all lines in multi-line comments.

notes-with-indexeddb/background.js
```diff
@@ -0,0 +1,14 @@
...
+*/
+function openNotes() {
+    chrome.tabs.create({
+        "url": chrome.extension.getURL("notes.html")
```
kmaglione: The `chrome.extension` namespace is being phased out, so please use `runtime.getURL` instead.

notes-with-indexeddb/background.js
```diff
@@ -0,0 +1,14 @@
+/*
+Open a new tab, and load "notes.html" into it.
+*/
+function openNotes() {
```
kmaglione: It would be nice to query whether a matching tab already exists and switch to it if it does, but I'm happy to leave that for a follow-up.

notes-with-indexeddb/notes.js
```diff
@@ -0,0 +1,169 @@
...
+
+if (!window.indexedDB) {
+    window.alert("Your browser doesn't support a stable version of IndexedDB.");
+}
+
+/*
+showErrorMessage from event if available
+ */
+function showErrorMessage(event) {
+    if (event && event.target && event.target.errorCode) {
```
kmaglione: No need for the `event && event.target` checks. Those should always be defined.

notes-with-indexeddb/notes.js
```diff
@@ -0,0 +1,169 @@
...
+
+window.IDBKeyRange = window.IDBKeyRange || window.webkitIDBKeyRange || window.msIDBKeyRange;
+
+if (!window.indexedDB) {
+    window.alert("Your browser doesn't support a stable version of IndexedDB.");
+}
+
+/*
+showErrorMessage from event if available
+ */
```
kmaglione: It would be nice to have complete [JSDoc](http://usejsdoc.org/about-getting-started.html) comments for all of these functions, but I won't require it.

notes-with-indexeddb/notes.js
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    alert('IndexedDB request denied');
+}
+
+/*
+addItem to db
+ */
+function addItem(db, data, success) {
+    var store = db.transaction(['notes'], 'readwrite').objectStore('notes'),
```
kmaglione: Nit: Please use `let` rather than `var`, and only declare one variable per `let` keyword. Using one `let` or `var` statement for declarations across multiple lines is a common cause of errors when refactoring.

The same goes for the other functions.

notes-with-indexeddb/notes.js
```diff
@@ -0,0 +1,169 @@
...
+        request = store.add(data);
+
+    request.onerror = showErrorMessage;
+    request.onsuccess = success;
```
kmaglione: Since most of our APIs and examples are moving toward using promises rather than callbacks, please have these functions do the same. You should be able to achieve it fairly easily with a helper function like this:

```javascript
function wrapRequest(request) {
  return new Promise((resolve, reject) => {
    request.onsuccess = resolve;
    request.onerror = reject;
  });
}
```

and then something like `return wrapRequest(store.add(data));` here.

notes-with-indexeddb/notes.js
```diff
@@ -0,0 +1,169 @@
...
+        var cursor = event.target.result;
+
+        if (cursor) {
+            var title = document.createElement('h3'),
+                description = document.createElement('p'),
+                item = document.createElement('div'),
+                button = document.createElement('button');
+
+            title.innerHTML = cursor.value.title;
+            description.innerHTML = cursor.value.description;
```
kmaglione: Please assign to `textContent` rather than `innerHTML` to avoid things accidentally being interpreted as markup.

notes-with-indexeddb/notes.js
```diff
@@ -0,0 +1,169 @@
...
+
+            item.appendChild(title);
+            item.appendChild(description);
+            item.appendChild(button);
+
+            list.appendChild(item);
+
+            cursor.continue();
+        } else {
+            notes.innerHTML = '';
```
kmaglione: Please also use `textContent` rather than `innerHTML` here.

notes-with-indexeddb/notes.js
```diff
@@ -0,0 +1,169 @@
...
+
+        addItem(db, {
+            title: title.value,
+            description: description.value
+        }, function() {
+            showNotes(db);
+            form.reset();
+        });
+
+        event.preventDefault();
```
kmaglione: Please move this to the start of the event handler so that if there's an unexpected exception, it still gets called.

notes-with-indexeddb/notes.js
```diff
@@ -0,0 +1,169 @@
...
+
+    var db = event.target.result;
+    db.onerror = showErrorMessage;
+
+    populateDb(db, seedData);
+    showNotes(db);
+
+    // add new item
+    var form = document.getElementById('add-new-item');
+    form.addEventListener('submit', function(event) {
```
kmaglione: Nit: Please use [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) for anonymous functions.

notes-with-indexeddb/notes.html
iampeterbanjo: ok - done

notes-with-indexeddb/notes.html
iampeterbanjo: ok

notes-with-indexeddb/notes.html
iampeterbanjo: ok - and removed in other files too

notes-with-indexeddb/manifest.json
iampeterbanjo: ok - and removed in other files too

notes-with-indexeddb/manifest.json
iampeterbanjo: yes. thanks

notes-with-indexeddb/manifest.json
iampeterbanjo: ok

notes-with-indexeddb/notes.js
iampeterbanjo: ok

notes-with-indexeddb/notes.js
iampeterbanjo: updated to JSDoc style

notes-with-indexeddb/notes.js
iampeterbanjo: ok

notes-with-indexeddb/notes.js
iampeterbanjo: ok - done

notes-with-indexeddb/notes.js
iampeterbanjo: thanks for the tip

notes-with-indexeddb/notes.js
iampeterbanjo: ok

notes-with-indexeddb/notes.js
iampeterbanjo: no problem

notes-with-indexeddb/notes.js
iampeterbanjo: ok

notes-with-indexeddb/notes.js
iampeterbanjo: alright

notes-with-indexeddb/background.js
iampeterbanjo: done

notes-with-indexeddb/background.js
iampeterbanjo: ok

notes-with-indexeddb/background.js
iampeterbanjo: alright - I'm not sure how to do this. the only unique thing about the tab is its url but I don't know how that's generated or if that's the best approach to look for a match.

notes-with-indexeddb/notes.html
iampeterbanjo: should we add a source code formatting config like jsbeautifyrc or similar and define the styles upfront?

wbamberg:
Closing: there's not been activity on this for a little while, and I'm not sure right now which storage examples we should have here.

Issue #122, opened 2016-10-25
Create an example that includes tests and how tor un them.


sblask:
Something like this: https://developer.mozilla.org/en-US/Add-ons/SDK/Tutorials/Unit_testing but for webextensions.

kumar303:
Thank you @hmelenok for creating this example! https://github.com/mdn/webextensions-examples/pull/130

PR #130, opened 2016-10-30
Create an example that includes tests and how to run them. #122

hmelenok:
Please check #122


wbamberg:
@hmelenok , nice work!

@andymckay , do you know who might be qualified to review this? There's a lot of stuff in here I don't know, so while I could probably figure out whether it's working properly, I probably won't be able to tell whether it's good practice.

andymckay:
@kumar303 might be able to help.

mocha-client-tests/addon/manifest.json
```diff
@@ -0,0 +1,53 @@
...
+  "browser_action": {
+    "default_icon": {
+      "19": "images/icon-19.png"
+    },
+    "default_title": "Mocha Test",
+    "default_popup": "popup.html"
+  },
+  "applications": {
+    "gecko": {
+      "id": "production@gemshelf.com",
```
andymckay: We've been trying to remove the id out of example extensions, they can be through temporary install locally or pushed to AMO for signing. If there's no id, they won't all conflict with each other.

hmelenok: Copy paste mistake

hmelenok: Please check it

mocha-client-tests/tests/lib/background.test.js
```diff
@@ -0,0 +1,9 @@
+describe('Background', function() {
+    describe('ping', function() {
+        it('should return pong in response', function() {
+            Background.ping(false, false, function(response) {
+                response.should.equal('pong');
```
kumar303: chai's prototype hack might be confusing to the casual JavaScript reader. Could you update this to use [expect()](http://chaijs.com/api/bdd/) (or something) instead? It would look like `expect(response).to.equal('pong')`

mocha-client-tests/addon/tests/lib/background-messaging.test.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Created by hmelenok on 10/30/16.
+ */
+describe('Background', function() {
+    describe('ping', function() {
+        it('should return pong in response', function(done) {
```
kumar303: how do I run this test? It isn't executed with `npm run test` from the root directory. Is there a different way to run it?

mocha-client-tests/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+And inside of addon please run `npm i`
+
+##Run with web-ext cli
+If you have it already run `npm run web-ext` if not `npm i web-ext -g` and then first command (will work with FF dev edition), if you have error with web-ext cli please add path for FF binary file with `--firefox-binary /path/to/firefox-bin`
+[web-ext docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference)
+##PhantomJs tests
+`npm run test` will run simple test  `./addon/background.js`
+## Mocha browser test
+Already inside of addon
```
kumar303: What do you mean by this? I couldn't figure out how to run the tests in `addon/tests/lib`. How do I run them and see results?

mocha-client-tests/addon/package.json
```diff
@@ -0,0 +1,16 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "MPL-2.0",
+  "devDependencies": {
+    "expect.js": "^0.3.1",
+    "mocha": "^3.1.2",
+    "should": "^11.1.1"
+  }
+}
```
hmelenok: There two type of tests inside of addon i had add mocha client tests

hmelenok: Outside of addon folder there phantomjs tests

mocha-client-tests/addon/scripts/popup.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Created by hmelenok on 10/30/16.
```
hmelenok: Done removed

mocha-client-tests/addon/tests/lib/background-messaging.test.js
hmelenok: `npm run test` - runs phantomjs tests, there UI ob mocha which included as example if you install all dependencies, please read README.md before

mocha-client-tests/README.md
hmelenok: `cd mocha-client-tests/addon && npm i` - it will install all node modules which required for running tests in addon window

hmelenok: Then just run `web-ext run  -s ./addon` with cli

hmelenok:
Please check last changes: 

- removed comments from js files
- removed jquery and empty contentscript 

screen shot 2016-11-05 at 19 42 53

mocha-client-tests/tests/lib/background.test.js
hmelenok: Thanks, changed to expect 👍

mocha-client-tests/addon/tests/lib/test.array.js
```diff
@@ -0,0 +1,8 @@
+describe('Array', function() {
+    describe('#indexOf()', function() {
+        it('should return -1 when the value is not present', function() {
+            [1,2,3].indexOf(5).should.equal(-1);
```
kumar303: This fails with `should is undefined`; it needs to use `expect()` instead, right?

screenshot 2016-11-07 10 59 25

mocha-client-tests/README.md
```diff
@@ -0,0 +1,13 @@
+#Mocha client tests for WebExtensions
+##Install dependency
+`npm install` and `bower install`  
+
+And inside of addon please run `npm i`
+
+##Run with web-ext cli
+If you have it already run `npm run web-ext` if not `npm i web-ext -g` and then first command (will work with FF dev edition), if you have error with web-ext cli please add path for FF binary file with `--firefox-binary /path/to/firefox-bin`
```
kumar303: `npm run web-ext` doesn't work because `package.json` wants it to be `npm run web-ext-run`. I think you should rename the `script` in `package.json` to `web-ext`.

kumar303: `npm install -g web-ext` is not necessary because you've already put it in the `devDependencies`. When you do `npm run` it figures out the path to `web-ext` automatically.

mocha-client-tests/README.md
```diff
@@ -0,0 +1,13 @@
+#Mocha client tests for WebExtensions
+##Install dependency
+`npm install` and `bower install`  
+
+And inside of addon please run `npm i`
```
kumar303: I think it would be more helpful to spell out `npm install` in all cases since not all readers will know about the `i` alias.

mocha-client-tests/README.md
kumar303: These instructions need a final sentence explaining that you need to click the mocha icon from the browser bar to run the tests

mocha-client-tests/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+##Run with web-ext cli
+If you have it already run `npm run web-ext` if not `npm i web-ext -g` and then first command (will work with FF dev edition), if you have error with web-ext cli please add path for FF binary file with `--firefox-binary /path/to/firefox-bin`
+[web-ext docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference)
+##PhantomJs tests
+`npm run test` will run simple test  `./addon/background.js`
```
kumar303: This can just be `npm test` which is a more standard way to run it using `npm`.

mocha-client-tests/README.md
kumar303: This "Mocha browser test" section is redundant, right? If so, I suggest removing it. It seems to be identical to "Run with web-ext cli."

mocha-client-tests/addon/tests/lib/background-messaging.test.js
```diff
@@ -0,0 +1,10 @@
...
+            chrome.runtime.sendMessage({action: 'ping'},function(response) {
```
kumar303: to keep the style consistent, add a space before `function`

mocha-client-tests/addon/background.js
```diff
@@ -0,0 +1,13 @@
+var Background = {
+    receiveMessage: function(msg, sender, sendResponse) {
+        if (msg && msg.action && Background.hasOwnProperty(msg.action)) {
+            return Background[msg.action](msg, sender, sendResponse);
+        }
```
kumar303: it would be nice if this had an else clause to warn about missed messages. It could help someone who may be editing this code. Something like:
````
... else {
    console.warning('No handler for message: ' + JSON.stringify(msg));
}
````

File: mocha-client-tests/addon/tests/lib/background-messaging.test.js [hunk 13]
```diff
@@ -0,0 +1,10 @@
...
+                expect(response).to.equal('pong');
```
kumar303: when this assertion fails, it only gets logged to the browser console. Can you make it show up in the mocha test output instead? For example, if you change `expect(response).to.equal('pong')` to `expect(response).to.equal('nope')` then I see this confusing timeout instead of the assertion error:

screenshot 2016-11-07 11 25 33

mocha-client-tests/addon/manifest.json
```diff
@@ -0,0 +1,38 @@
...
+    "<all_urls>"
+  ],
+  "browser_action": {
+    "default_icon": {
+      "19": "images/icon-19.png"
+    },
+    "default_title": "Mocha Test",
+    "default_popup": "popup.html"
+  },
+  "applications": {
```
kumar303: You can remove this section entirely. It's not needed for this example. It could be harmful if someone copy/pasted it because it makes the extension only compatible with Firefox, not Chrome.

mocha-client-tests/addon/manifest.json
```diff
@@ -0,0 +1,38 @@
...
+  "icons": {
+    "16": "images/icon-16.png"
+  },
+  "short_name": "MochaTest",
+  "background": {
+    "scripts": [
+      "background.js"
+    ]
+  },
+  "permissions": [
```
kumar303: This looks like it has a bunch of permissions not needed for this example. Can you strip it down to only the relevant permissions? I think only `notifications` is needed, right?

mocha-client-tests/addon/tests/mocha-run.js
```diff
@@ -0,0 +1,3 @@
+mocha.checkLeaks();
+mocha.globals(['jQuery','AppView', 'ExtensionOptions', 'ExtensionView', 'WebView']);
```
kumar303: where are these globals used? These don't look like they are relevant to the example

mocha-client-tests/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "main": "index.js",
+  "scripts": {
+    "test": "mocha-phantomjs -p ./node_modules/phantomjs-prebuilt/bin/phantomjs -R nyan tests/background.html",
+    "web-ext-run": "web-ext run  -s ./addon"
+  },
+  "author": "",
+  "license": "MPL-2.0",
+  "devDependencies": {
+    "chai": "^3.5.0",
+    "chrome-mock": "0.0.9",
```
kumar303: where is this used?

mocha-client-tests/addon/tests/mocha-run.js
hmelenok: There a list of default globals in extension which from Mocha perspective looks like global leaks

mocha-client-tests/package.json
hmelenok: chrome-mock used for phantomJS as it not have WebExtensions (or Chrome Extensions) API inside of engine, so i'm using mock of this API's. As when you include to test runner original file it throws errors about chrome is not defined

mocha-client-tests/README.md
hmelenok: Done

mocha-client-tests/README.md
hmelenok: Done

mocha-client-tests/addon/manifest.json
hmelenok: done

mocha-client-tests/addon/tests/lib/background-messaging.test.js
hmelenok: fixed

mocha-client-tests/addon/tests/lib/background-messaging.test.js
hmelenok: I'm not sure by which style (as i don't have spaces anywhere)

mocha-client-tests/addon/tests/lib/background-messaging.test.js
hmelenok: https://github.com/mochajs/mocha/issues/1128 - looks like I'm not sure why it wait until fail

mocha-client-tests/package.json
kumar303: aha, I see

kumar303:
I see that you pushed some more changes (thanks!). Just let me know when it's ready for another review.

mocha-client-tests/addon/tests/lib/background-messaging.test.js
kumar303: ok, I see what the issue is now. The `expect()` call throws an error but that is not passed to `done(error)` so the mocha reporter can't show it. The way to fix this is to rewrite the code using promises:
````javascript
it('should return pong in response', function() {
  // Return a promise for Mocha using the Firefox browser API instead of chrome.
  return browser.runtime.sendMessage({action: 'ping'})
    .then(function(response) {
      expect(response).to.equal('pong');
    });
});
````
I think it would be good to convert this code into promises since the developer can always use a [polyfill](https://github.com/mozilla/webextension-polyfill) to make it compatible with Chrome.

Comment (hmelenok) [inline mocha-client-tests/addon/tests/lib/background-messaging.test.js]: Thanks for advice it works!

Comment (hmelenok) [inline mocha-client-tests/addon/tests/lib/background-messaging.test.js]: Added in last commit

**Comment by hmelenok** [comment]
Please check it again)

File: mocha-client-tests/README.md [hunk 18]
```diff
@@ -0,0 +1,17 @@
...
+
+Than please run `cd ./addon/` and `npm install` to install mocha. It give you possibility to run client test inside of addon with mocha UI. If you don't want to have mocha UI you can install [WebConsole-reporter](https://github.com/eeroan/WebConsole-reporter)
+
+##Run with web-ext cli
+Just run `npm run web-ext` (will work with FF dev edition), if you have error with web-ext cli please add path for FF binary file with `--firefox-binary /path/to/firefox-bin`
+[(web-ext docs)](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference).
+
+When addon will start click on mocha icon in your browser bar to run client tests: 
+
+![addon screenshot](screenshots/addon-button.png "Mocha test addon")
```
kumar303: good idea to add a screenshot

mocha-client-tests/tests/lib/background.test.js
kumar303: I don't see the change to make this `expect()`. Do you have any unpushed commits maybe?

mocha-client-tests/addon/scripts/browser-polyfill.js
```diff
@@ -0,0 +1,849 @@
+/* webextension-polyfill - v0.1.0 - Wed Nov 09 2016 16:37:59 */
```
kumar303: this is the built dist file, right? I don't know what to look for.

hmelenok: That what i get after build in `webextension-polyfill`

mocha-client-tests/README.md
hmelenok: 👍

kumar303:
Thanks for all the fixups!

PR #131, opened 2016-10-31
Example of Chrome/Firefox compatible extension with polyfill

jeffesquivels:


jeffesquivels:
@kumar303 @wbamberg this is a first attempt to  use promises instead of callbacks, it is currently working on FF, but I get this error on Chromiun (version 53.0.2785.143):

screenshot from 2016-10-31 01-44-53

I'm sure there's something I'm missing here (I also researched and used the arrow function expression first, as it was used a lot in the examples I read first about promises, but when that failed too, I reverted to the simplest example I could think of that still used promises instead of callbacks), could you point me in the right direction?

Thanks,

kumar303:
@kmaglione does this Chrome error look like a bug we should file in https://github.com/mozilla/webextension-polyfill?

kmaglione:
The problem is that you used the source file rather than building and using the final build in `dist/browser-polyfill.js`

kumar303:
@kmaglione d'oh, I missed that. Thanks for taking a closer look.

@jeffesquivels this explains how to build a version of the polyfill that you can safely include: https://github.com/mozilla/webextension-polyfill#building

jeffesquivels:
@kmaglione thanks for the clarification, I'm a JS rookie and dind't realized there were source js files that needed to be built :).

@kumar303 thanks for the link, I'm installing npm right now.

BTW, thinking about forward compatibility, what would be the correct way to include the polyfill so that we're always using the latest version?

I was thinking using git submodules, but if there's a building phase then that wouldn't work automagically for people. Or would it make more sense (because we're just providing examples) that we just manually update the file in the repo now and then, what do you think?

Thanks.

kumar303:
> what would be the correct way to include the polyfill so that we're always using the latest version?

Good question. I think your current approach is the best way but you could add a comment in the readme warning that polyfill upgrades will need to be done manually. I filed this bug to address polyfill upgrades better: https://github.com/mozilla/webextension-polyfill/issues/7

> I was thinking using git submodules

Those are kind of painful in my experience so I don't think it's worth it.

wbamberg:
Closing, as this PR hasn't seen any activity in a while.

PR #132, opened 2016-10-31
remove window.parent check from contentscript

ally104:
The manifest 'all_frames' property defaults to false, so the content script should not need to perform the window.parent check.
To perform parent check they needs to be maintained with back version also.


wbamberg:
We decided to remove the inpage-toolbar-ui example, so I don't think this is needed.

PR #134, opened 2016-11-01
Duplicate example using the browser polyfill

jeffesquivels:
This commit creates a second version of the tabs, tabs, tabs example but
using the WebExtension browser API polyfill to demonstrate the possibility
to use the same extension on Firefox and Google Chrome without any
modifications to the source code.

### DIFF
--- a/tabs-tabs-tabs-v2/README.md
+++ b/tabs-tabs-tabs-v2/README.md
@@ -1,0 +1,11 @@
+# tabs, tabs, tabs
+
+## What it does
+
+This extension includes a browser action with a popup specified as "tabs.html".
+
+The popup lets the user perform various simple operations using the tabs API.
+
+# What it shows
+
+Demonstration of various tabs API functions.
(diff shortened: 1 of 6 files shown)


jeffesquivels:
@wbamberg as promised, this is the PR of the [tabs, tabs, tabs](https://github.com/mdn/webextensions-examples/tree/master/tabs-tabs-tabs) example already in the repo converted to use the browser polyfill.

I didn't convert it to using Promises yet as I ran into the same error I found with the other PR (#131), so I'm waiting for that issue to clear before continuing the conversion here.

Any comments/feedback are very much welcome.

wbamberg:
Hi @jeffesquivels 

I'm really sorry not to get back to you sooner about this :(. I thought it was better to get the polyfill working properly with promises first.

I just opened a PR to convert *all the examples* to use browser.* and promises (https://github.com/mdn/webextensions-examples/pull/138). 

So I think perhaps the best plan here would be to wait until that PR lands, then add the polyfill to one of the existing examples (such as tabs-tabs-tabs) so it works in Chrome. (That is, rather than duplicate tabs-tabs-tab, make it Chrome-compatible.) Then we can document that as an example to follow.

What do you think?

wbamberg:
That PR has now landed \o/.

wbamberg:
Closing, as this PR hasn't seen any activity in a while. We do still need an example using the polyfill though.

PR #135, opened 2016-11-01
add tabs-tabs-tabs example to show switching active tabs

iampeterbanjo:
Based on [feedback](https://github.com/mdn/webextensions-examples/pull/125) from @andymckay I'm opening this PR.

I noticed an issue with getting the tab title while working on FFx on Linux Mint - the title is undefined so I've added a fallback to tab id. Is this a known issue? Are there workarounds?


andymckay:
You could be hitting this bug?

https://bugzilla.mozilla.org/show_bug.cgi?id=1310019#c3

If not, we should file another.

tabs-tabs-tabs/tabs.css
```diff
@@ -7,6 +7,14 @@ a {
   display: inline-block;
 }
 
+.switch-tabs {
+	padding-left: 10px;
+}
+
+.switch-tabs a {
+	display: block;
```
andymckay: a minor nit, the indentation is different from the lines above it (eg. line 7)

tabs-tabs-tabs/tabs.js
```diff
@@ -12,6 +12,30 @@ function firstUnpinnedTab(tabs) {
...
+    var tabsList = document.getElementById('tabs-list'),
+        currentTabs = document.createDocumentFragment();
+
+    tabsList.innerHTML = '';
+
+    for (var tab of tabs) {
+      if (!tab.active) {
+        var tabLink = document.createElement('a');
+
+        tabLink.innerHTML = tab.title || tab.id;
```
andymckay: Use of innerHTML is not recommended due to XSS and performance concerns, I think using tabLink.textContent will work here instead?

tabs-tabs-tabs/tabs.html
```diff
@@ -33,6 +33,13 @@
     <a href="#" id="tabs-default-zoom">Reset zoom</a><br>
     <a href="#" id="tabs-decrease-zoom">Zoom out</a>
 
+    <div class="panel-section-separator"></div>
+
+    <div class="switch-tabs">
+      <p>Switch tabs</p>
```
andymckay: The term switch to me means swap two things around, could just be me. I wonder if "Active tab" or "Switch to tab" would make it clearer.

tabs-tabs-tabs/tabs.js
```diff
@@ -12,6 +12,30 @@ function firstUnpinnedTab(tabs) {
   }
 }
 
+function listTabs() {
+  chrome.tabs.query({
+    currentWindow: true
+  }, function(tabs) {
+    var tabsList = document.getElementById('tabs-list'),
+        currentTabs = document.createDocumentFragment();
```
andymckay: I don't think we enforce this in a style guide, but could you make this `var currentTabs` please.

tabs-tabs-tabs/tabs.js
```diff
@@ -12,6 +12,30 @@ function firstUnpinnedTab(tabs) {
...
+        var tabLink = document.createElement('a');
+
+        tabLink.innerHTML = tab.title || tab.id;
+        tabLink.setAttribute('href', tab.id);
+        tabLink.classList.add('switch-tabs');
+        currentTabs.appendChild(tabLink);
+      }
+    }
+
+    tabsList.appendChild(currentTabs);
```
andymckay: This list is going to get huge if you've got like 100 tabs open. People do wierd things like that, I wonder if we should limit this down to just a few to stop something bad happening. Once we've demonstrated the point on a few tabs, there's no need to do it for them all. 

We did this on history-deleter for a similar reason.

tabs-tabs-tabs/tabs.js
```diff
@@ -12,6 +12,30 @@ function firstUnpinnedTab(tabs) {
...
+  chrome.tabs.query({
+    currentWindow: true
+  }, function(tabs) {
+    var tabsList = document.getElementById('tabs-list'),
+        currentTabs = document.createDocumentFragment();
+
+    tabsList.innerHTML = '';
+
+    for (var tab of tabs) {
+      if (!tab.active) {
```
andymckay: This confused me, because the number of listed tabs is always less than the number of tabs I have. Also because the popup persists as the tab switches this gets out of sync. I'm not sure what a nice solution is there though. Perhaps is we limit it to 5 or so (see other comments) this won't be too bad.

tabs-tabs-tabs/tabs.css
iampeterbanjo: ok - changed tabs to spaces

tabs-tabs-tabs/tabs.js
iampeterbanjo: fixed

tabs-tabs-tabs/tabs.js
iampeterbanjo: ok. limited to 5

tabs-tabs-tabs/tabs.html
iampeterbanjo: ok - changed to "Switch to tab"

tabs-tabs-tabs/tabs.js
iampeterbanjo: changed to `tabLink.textContent`

andymckay:
This looks good, sadly https://github.com/mdn/webextensions-examples/pull/138 landed, so it now conflicts 😞 , so you need to rebase on that. And then its good to go.

Thanks for the changes and sorry for the delay in reviewing.

iampeterbanjo:
fixed conflicts

andymckay:
Thanks @iampeterbanjo

PR #137, opened 2016-11-05
Typo fix

xfq:
The delete button in Quicknote should delete a note, not an extension.


wbamberg:
Thanks for the fix @xfq !

PR #138, opened 2016-11-11
Rewrite examples to use browser.* and promises

wbamberg:
This is mostly boilerplate, replacing `browser.` with `chrome.` and using promises instead of callbacks. I've made a few other changes for readability and consistency (arrow functions and template literals, basically).

There's one actual, unrelated, fix: https://github.com/mdn/webextensions-examples/blob/master/embedded-webextension-bootstrapped/step2-pure-webextension/popup.js#L3 should be `get` not `set`.

I've manually tested every add-on that I changed.

@andymckay , would you please review this, or find someone else to? :-)

### DIFF
--- a/apply-css/background.js
+++ b/apply-css/background.js
@@ -11,17 +11,18 @@ function toggleCSS(tab) {
 
   function gotTitle(title) {
     if (title === TITLE_APPLY) {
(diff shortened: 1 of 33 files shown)


wbamberg:
Thanks for the review @andymckay !

PR #141, opened 2016-11-16
identity example

mixedpuppy:


identity/background.js
```diff
@@ -0,0 +1,233 @@
+
+let providers = {
+  facebook: {
+    clientId: '1094807753900920',
+    clientSecret: '3d4af261312493c4397e8003c9139500',
```
andymckay: Is this ok to be in the repo? Maybe a comment about what developers should do if they were copying and pasting?

mixedpuppy: yeah, it's no big deal.  But should link to docs I suppose.  And kill off those apps in my account.

identity/manifest.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "chrome.identity example",
+  "version": "1",
+  "manifest_version": 2,
+  "applications": {
+    "gecko": { "id": "identity.example@mozilla.org" }
```
andymckay: We should set a minimum browser version here of 53? We've also been trying to remove the id, so it can load from web-ext and not block on AMO if you try to sign it.

identity/manifest.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "chrome.identity example",
```
andymckay: @wbamberg has standardized on using browser instead of chrome, so that should apply for all the APIs in this.

identity/manifest.json
mixedpuppy: Since the id changes every time you reload the temporary addon, and the redirect uri is based on the id, you kind of need a hard set id to test oauth stuff.

andymckay: Ah right. Could you add a comment about that somewhere in the code, that's a useful peice of information to anyone using the API and will stop someone removing that in the future 😄

identity/README.md
```diff
@@ -0,0 +1,49 @@
...
+# First time run before setting up applications.
+
+You will want to set a specific ID in your manifest.json file.  In this example,
+it is set to identity.example@mozilla.org.  The identity API has a helper
+function that generates a hashed URL based on the ID.  If you do not have a
+specific ID set, the extension id will change every time you install the
+extension (or reload as a temporary addon).
+
+Load the extension as a temporary addon in about:debugging then open the
+Brower console and look for a message: Your redirect_uri is XXX.  This message
```
andymckay: Ugh, I found this hard to find amongst the noise. Would it be possible to place on `index.html`? And we could put a few more of these docs there too? Basically just means adding something like:

`document.getElementById('redirect_url').innerText = browser.identity.getRedirectURL("callback");`

..to index.js

identity/background.js
```diff
@@ -0,0 +1,237 @@
...
+    user_info: null,
+
+    clear() {
+      this.user_info = null;
+      this.token = null;
+    },
+
+    getAuthURL(redirectURL) {
+      return 'https://www.facebook.com/dialog/oauth?client_id=' + this.clientId +
+             '&reponse_type=token&access_type=online&display=popup' +
```
andymckay: just to check, is `reponse` correct in `reponse_type`?

identity/index.js
```diff
@@ -0,0 +1,23 @@
...
+  chrome.runtime.sendMessage({"type": "getUserInfo", provider, "interactive": intractive});
+}
+
+function revoke() {
+  chrome.runtime.sendMessage({"type": "removeCachedToken"});
+  show();
+}
+
+function show(text) {
+  document.getElementById('user_info').innerHTML = text || "";
```
andymckay: Lets use innerText as good practice to avoid any XSS issues.

identity/index.js
```diff
@@ -0,0 +1,23 @@
...
+}
+
+function messageListener(message) {
+  show("<pre>" + JSON.stringify(message.error || message.user) + "</pre>");
```
andymckay: Couldn't you just put that the `<pre>` in the HTML document?

identity/background.js
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    let options = {
+      interactive: interactive,
+      url: provider.getAuthURL(redirectUri)
+    };
+
+    browser.identity.launchWebAuthFlow(options, function(redirectURL) {
+      if (chrome.runtime.lastError) {
+        reject(new Error(chrome.runtime.lastError.message));
```
andymckay: We've switched all the examples to `browser` now.

wbamberg:
I'm worried that "background.js" is quite a lot of quite complex (to me anyway) code. Would it be possible to add some comments about what the different functions are doing?

wbamberg:
Also, I'm not sure I understand the dance we do with redirect_url. I mean, I understand how to do it, but I don't understand how this would map onto a real-world add-on. Obviously we wouldn't expect real users of a real add-on to do something like this, so how would they use this API?

identity/index.js
```diff
@@ -0,0 +1,23 @@
+function getUserInfo(provider, intractive=true) {
```
wbamberg: -> interactive

identity/background.js
```diff
@@ -0,0 +1,237 @@
...
+      return new Promise((resolve, reject) => {
+        if (params.get("error")) {
+          reject(new Error(params.get("error_description")));
+        } else if (params.get("id_token")) {
+          this.token = params.get("id_token");
+          resolve(this.token);
+        } else if (params.get("access_token")) {
+          let token = params.get("access_token");
+          this.verifyToken(token).then(verified => {
+            // TODO verify the data
```
wbamberg: Is this TODO important?

identity/background.js
wbamberg: (and to use promises instead of callbacks)

kumar303: > We've switched all the examples to browser now.

Wait, really? That means the examples won't run in Chrome though, right? That seems excessively harsh, especially if we want to encourage devs to write cross-platform extensions.

wbamberg: Yes. This was discussed on list: https://mail.mozilla.org/pipermail/dev-addons/2016-October/002025.html.

andymckay:
With this landing in 53 and being merged, I wondered if we could get this landed?

/cc @mixedpuppy, @wbamberg

wbamberg:
Closing this now https://github.com/mdn/webextensions-examples/pull/180 has landed.

PR #144, opened 2016-11-17
Remove applications key wherever possible

wbamberg:
Fix for https://github.com/mdn/webextensions-examples/issues/71, following the plan from https://github.com/mdn/webextensions-examples/issues/71#issuecomment-253642541, but also taking into account @andymckay 's https://github.com/mdn/webextensions-examples/issues/71#issuecomment-253645244.

I've removed the applications key in all cases except:
* the embedded-* examples, since they need and explicit ID and a newer-than-released Firefox
* native-messaging, for the same reasons (actually this requires Firefox 50, which is the release, but only just...)
* navigation-stats, since it needs Firefox 50, which is only just the release

### DIFF
--- a/selection-to-clipboard/manifest.json
+++ b/selection-to-clipboard/manifest.json
@@ -9,12 +9,6 @@
         "48": "icons/clipboard-48.png"
     },
 
-    "applications": {
-        "gecko": {
-            "strict_min_version": "51.0a1"
-        }
-    },
-
     "content_scripts": [{
         "matches": ["<all_urls>"],
         "js": ["content-script.js"]
--- a/apply-css/manifest.json
+++ b/apply-css/manifest.json
@@ -6,13 +6,6 @@
   "version": "1.0",
   "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/apply-css",
 
-  "applications": {
-    "gecko": {
-      "id": "apply-css@mozilla.org",
-      "strict_min_version": "49.0"
-    }
-  },
-
   "background": {
     "scripts": ["background.js"]
   },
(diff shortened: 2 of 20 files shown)


wbamberg:
@kumar303 , would you mind taking a look at this, please?

kumar303:
sure, I can look

selection-to-clipboard/manifest.json
```diff
@@ -9,14 +9,8 @@
         "48": "icons/clipboard-48.png"
     },
 
-    "applications": {
-        "gecko": {
-            "strict_min_version": "51.0a1"
```
kumar303: Did this really require 51 or not?

kumar303:
What about `navigation-stats`? I see that it currently requires Firefox 50 but I couldn't see why. Which API does it need?

selection-to-clipboard/manifest.json
wbamberg: No. The original PR asked for the "clipboardWrite" permission, that was added in 51. But the example doesn't actually need that permission, because the copy is done in an event handler for a user action. So I asked the contributor to remove the permission, so as not to suggest to people that it is in fact needed.

The example works fine in Firefox 50.

wbamberg:
navigation-stats needs Firefox 50, I think, because it passes a filter to webNavigation.onCompleted (https://github.com/mdn/webextensions-examples/blob/master/navigation-stats/background.js#L29), and this API only supports filtering from Firefox 50: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webNavigation/onCompleted#Browser_compatibility.

wbamberg:
Thanks for the review @kumar303 !

PR #149, opened 2016-12-01
Modify mocha-test-clients to use Karma

sneakypete81:
Using PhantomJS for testing WebExtensions is not ideal. It uses webkit under the hood, which can have implementation differences and missing features. It doesn't support ES2015 classes, for example.

This PR:
* Modifies the example to use Karma to run the commandline tests in a Firefox instance.
* Switches from chome-mock to sinon-chrome, which [supports the WebExtensions API](https://github.com/acvetkov/sinon-chrome/issues/40).
* Updates the README to more clearly explain the distinction between in-addon and commandline testing, and add a link to the excellent https://github.com/Standard8/example-webextension.


kumar303:
Thanks so much! I didn't want to hold up the other patch but I also wanted to see it using a real browser too. High five. I'll take a closer look.

kumar303:
This isn't related to your patch but I noticed that the background test doesn't handle async messaging correctly. Can you apply this patch while you're in here?
````diff
diff --git a/mocha-client-tests/tests/lib/background.test.js b/mocha-client-tests/tests/lib/background.test.js
index bf42c31..0588bb3 100644
--- a/mocha-client-tests/tests/lib/background.test.js
+++ b/mocha-client-tests/tests/lib/background.test.js
@@ -1,9 +1,10 @@
 describe('Background', function() {
     describe('ping', function() {
-        it('should return pong in response', function() {
+        it('should return pong in response', function(done) {
             Background.ping(false, false, function(response) {
                 expect(response).to.equal('pong');
+                done();
             });
         });
     });
-});
\ No newline at end of file
+});
````

File: mocha-client-tests/README.md [hunk 1]
```diff
@@ -1,17 +1,39 @@
 #Mocha client tests for WebExtensions
-##Install dependency
-`npm install` - will install all dependencies for running PhantomJS outside of addon
+##Introduction
+This example shows two methods of testing a WebExtension:
+* Running tests from within the addon
+* Running tests from the commandline using Karma
 
-Than please run `cd ./addon/` and `npm install` to install mocha. It give you possibility to run client test inside of addon with mocha UI. If you don't want to have mocha UI you can install [WebConsole-reporter](https://github.com/eeroan/WebConsole-reporter)
+See https://github.com/Standard8/example-webextension for a more complete example of WebExtension test configuration.
```
kumar303: 👍

mocha-client-tests/README.md
```diff
@@ -1,17 +1,39 @@
...
+  npm install
+```
+To run tests from within the addon:
+```
+  cd addon
+  npm install
+```
+
+##Testing within the Addon
+This gives you the possibility to run client test inside the addon with the mocha UI.
```
kumar303: typo: it should be: ...possibility to run client tests...

mocha-client-tests/README.md
```diff
@@ -1,17 +1,39 @@
...
+If you don't want to use the mocha UI, you can install [WebConsole-reporter](https://github.com/eeroan/WebConsole-reporter).
+
+###Run with web-ext cli
 Just run `npm run web-ext` (will work with FF dev edition), if you have error with web-ext cli please add path for FF binary file with `--firefox-binary /path/to/firefox-bin`
 [(web-ext docs)](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference).
 
-When addon will start click on mocha icon in your browser bar to run client tests: 
+When the addon starts, click on mocha icon in your browser bar to run client tests:
```
kumar303: it should be: ... click on the mocha icon...

mocha-client-tests/README.md
```diff
@@ -1,17 +1,39 @@
...
\ No newline at end of file
+This will test  `./addon/background.js` with `./addon/tests/lib/background-messaging.test.js`.
+
+##Testing from the Commandline
+This uses [Karma](http://karma-runner.github.io) to run tests from the commandline. Just type `npm test` to test `./addon/background.js` with `./tests/lib/background.test.js`.
+
+###Debug  Mode
+Use `npm run test:debug` to run Karma in watch mode. Whenever you modify a Javascript file, the tests will automatically rerun.
+
+You can install [karma-notification-reporter](https://www.npmjs.com/package/karma-notification-reporter) to display test results in a desktop notification.
```
kumar303: This looks like it also requires configuration, right? Can you document how to configure it? I tried installing it but I didn't get any notifications of test failures.

mocha-client-tests/karma.conf.js
```diff
@@ -0,0 +1,72 @@
...
+
+    // frameworks to use
+    // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
+    frameworks: ['mocha'],
+
+    // list of files / patterns to load in the browser
+    files: [
+      // Test dependencies
+      'node_modules/expect.js/index.js',
+      'node_modules/sinon-chrome/bundle/sinon-chrome-webextensions.min.js',
```
kumar303: can't we move all of the hard coded dependency scripts into the config now? https://github.com/sneakypete81/webextensions-examples/blob/8e691c91af224d3880f41d7d98f721b211d9d6e4/mocha-client-tests/tests/background.html#L10-L13

kumar303: is it possible to use a non minified version of `sinon-chrome-webextensions.min.js` for debugging?

kumar303: actually, I'm a little confused about `background.html` -- do we still need it now that it's running from karma?

mocha-client-tests/karma.conf.js
```diff
@@ -0,0 +1,72 @@
...
+
+      // Source
+      'addon/*.js',
+
+      // Tests
+      'tests/lib/*.js'
+    ],
+
+    // list of files to exclude
```
kumar303: since this is a little confusing, I would suggest adding a note like "These tests are intended to be run from the add-on itself, not from the karma test suite"

kumar303:
oops, I noticed something else unrelated to your patch. Could you fix this? It would be nice to include a test to cover this but you'd have to create a separate message object so I wouldn't worry about it.

````diff
diff --git a/mocha-client-tests/addon/background.js b/mocha-client-tests/addon/background.js
index 1bda77c..c434fcc 100644
--- a/mocha-client-tests/addon/background.js
+++ b/mocha-client-tests/addon/background.js
@@ -3,7 +3,7 @@ var Background = {
         if (msg && msg.action && Background.hasOwnProperty(msg.action)) {
             return Background[msg.action](msg, sender, sendResponse);
         } else {
-            console.warning('No handler for message: ' + JSON.stringify(msg));
+            console.warn('No handler for message: ' + JSON.stringify(msg));
         }
     },
     ping: function(msg, sender, sendResponse) {
@@ -12,4 +12,4 @@ var Background = {
     }
 };
 
-chrome.runtime.onMessage.addListener(Background.receiveMessage);
\ No newline at end of file
+chrome.runtime.onMessage.addListener(Background.receiveMessage);
````

mocha-client-tests/karma.conf.js
sneakypete81: You're right, Karma generates its own html testrunner. I'll delete background.html.

sneakypete81:
Thanks @kumar303, changes applied.

mocha-client-tests/karma.conf.js
sneakypete81: BTW @kumar303, slightly related: do you know if the WebExtensions schemas being kept up to date for the browser namespace?

I recently tried to use one (storage.StorageArea.get) and found it defined to use a callback rather than a promise:
https://hg.mozilla.org/mozilla-central/file/tip/toolkit/components/extensions/schemas/storage.json
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/storage/StorageArea/get

kumar303: The schemas are kept up to date but the one you linked to is the schema for a Chrome compatible interface, the one documented here: https://developer.chrome.com/extensions/storage#type-StorageArea

The schema file that describes Mozilla's promise based interface is somewhere else but I can't find it at the moment! @rpl works with them a lot so maybe he can find the exact place in mozilla central.

rpl: @sneakypete81 when in the WebExtensions API schema files there is a `"async"` property on a method definition with a callback parameter, the callback parameter can be omitted and that API method will return a Promise, e.g. for the `StorageArea.get` is at line https://hg.mozilla.org/mozilla-central/file/tip/toolkit/components/extensions/schemas/storage.json#l37

(the value of the `"async"` property is the name of the callback parameter, that is almost always `"callback"`)

sneakypete81: @kumar303 @rpl: thanks to you both, that makes sense. I'll take a quick look at adding Promise support to sinon-chrome for methods with async properties.

sneakypete81: :blush: (ignore that last comment - I'd misunderstood what sinon-chrome was doing. It works fine with both callback and Promise variants).

PR #155, opened 2016-12-09
Example using React / ES6

kumar303:
This is an example that uses React and ES6 for a popup UI


kumar303:
@rpl @wbamberg could you take a look at this? I think it will be helpful since React is quite popular for building UIs.

react-es6-popup/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build": "webpack -w -v --display-error-details --progress --colors",
+    "start": "web-ext run -s extension/"
+  },
+  "author": "",
+  "license": "MPL-2.0",
+  "devDependencies": {
+    "babel-core": "^6.20.0",
```
rpl: Nit: there aren't a lot of npm dependencies and they are probably following the semver convention really well, and so it is probably not a big issue, but it could make sense to use exact versions for the npm dependencies so that we are absolutely sure that the example will not break accidentally because of potential incompatible changes in a minor version update of these dependencies.

react-es6-popup/src/popup.js
```diff
@@ -0,0 +1,20 @@
...
+        <p>
+          This is an example of a popup UI in React.
+        </p>
+        <Nested />
+      </div>
+    );
+  }
+}
+
+ReactDOM.render(<Popup/>, document.getElementById('app'));
```
rpl: it would be nice to make a single WebExtensions API call here and render the component with some data in it (without add too much complexity to the example, e.g. just a single call to `browser.tabs.query` or `browser.storage.local.get/set`)

react-es6-popup/webpack.config.js
```diff
@@ -0,0 +1,48 @@
...
+    loaders: [{
+      exclude: /node_modules/,
+      test: /\.js$/,
+      // Babel options are in .babelrc
+      loaders: ['babel'],
+    }],
+  },
+  resolve: {
+    // This allows you to import modules just like you would in a NodeJS app.
+    extensions: ['', '.js', '.jsx'],
```
rpl: Why there is an empty (`''`) extension here? is it needed?

wbamberg:
> it could even make sense to add a tutorial version for it on MDN

-> https://bugzilla.mozilla.org/show_bug.cgi?id=1322750 so I don't forget.

react-es6-popup/webpack.config.js
kumar303: I'm not totally sure why but if I remove it, webpack gives me this error:

ERROR in ./~/react-dom/lib/flattenChildren.js
Module not found: Error: Cannot resolve 'file' or 'directory' ./../../process/browser.js in /Users/kumar/dev/webextensions-examples/react-es6-popup/node_modules/react-dom/lib

kumar303: oh, huh, part of the error included this as one of the paths it tried. I guess this is how it builds module paths:

/Users/kumar/dev/webextensions-examples/react-es6-popup/node_modules/process/browser.js.js

react-es6-popup/src/popup.js
kumar303: Oh yeah, good idea. I added an example calling an API.

kumar303:
@rpl thanks for the detailed review, I fixed everything up. @wbamberg sorry I missed you at mozaloha! Let me know if there's anything else this needs and I'll do a follow-up commit.

wbamberg:
@kumar303 , sure thing! Sorry I didn't get a chance to look at it properly. I'm sure between you and @rpl it will be great :).

Issue #156, opened 2016-12-11
Native messaging examples fail with python3


evolighting:
Notification about this issue is need, i propose.

wbamberg:
Fixed by https://github.com/mdn/webextensions-examples/pull/157.

PR #157, opened 2016-12-12
Fix native message examples with python3

evolighting:
I/O behave differently between 2 and 3


wbamberg:
(Note, for more context: this is a fix for https://github.com/mdn/webextensions-examples/issues/156).

Thanks! This makes sense, and seems to work great in Python 3. I wondered if it might be possible to get the same code to work in both 2 and 3? I had a quick look and couldn't see anything, but I'm no expert (it seems like it's basically a matter of being able to read/write raw bytes from/to stdin?).

CC @aswan, whose original Python code I stole for this :).

andymckay:
It should be a matter of just detecting if sys.stdout.buffer is available and then doing a couple of `if` statements based on that and one file would work in both python 2 and 3.

I would also be awesome if we pep-8'd this code, I'll do that in a seperate pull request.

wbamberg:
@andymckay , I have had a go at doing this - would you mind taking a look please? It's tested and appears to work with Pythons 2 and 3.

native-messaging/app/ping_pong.py
```diff
@@ -1,34 +1,62 @@
...
+        sys.stdout.write(encodedMessage['content'])
+        sys.stdout.flush()
 
-while True:
-    receivedMessage = getMessage()
-    if receivedMessage == 'ping':
-        sendMessage(encodeMessage('pong'))
+    while True:
+        receivedMessage = getMessage()
+        if (receivedMessage == "ping"):
```
andymckay: nit: i don't think parentheses are needed in python 2 or 3

wbamberg:
Thanks!

PR #159, opened 2016-12-12
make python pep 8 compliant

andymckay:


andymckay:
That gif totally makes up for it and then some.

PR #162, opened 2016-12-12
add in some built versions

andymckay:
* a first pass at building the extensions so people don't have to


wbamberg:
Is the idea that we rebuild-and-sign these manually when they change, or did you have some clever way to automate that? If the former, did you sign this using some particular AMO account that maintainers need the credentials for?

andymckay:
I signed them using my account. I was thinking we give a few people (like you) owner access so we can also sign them. I started writing a command to automate it and could do that, but all it does is run through and do web-ext sign on each one...

wbamberg:
OK. Do you think we should have a dedicated account for these examples on AMO, or is it OK to use our personal accounts?

andymckay:
Done, you should have about 4000 emails, sorry about that.

PR #163, opened 2016-12-13
fixed manifest.json of context-menu-demo to make it comply with il8n

### LINKED ISSUE #148: context-menu-demo doubles as i18n demo, but missing manifest.json entries
In context-menu-demo, it is great to see an example of how to use i18n in a background script.

However, as this doubles as an [i18n demo](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/i18n#Example_add-ons), I think the manifest.json should also show proper i18n usage.

manifest.json should remove the existing _name_ field, and replace it with the following.  Note that the __locales/en/messages.json_ file already has the required entries.

```
"name": "__MSG_extensionName__",
"description": "__MSG_extensionDescription__",
```

This is a powerful and simple feature that was not possible in XUL/Overlays, and cumbersome (if even possible) in SDK.  But WebExtensions has it very cleanly implemented.  A definite selling point for authors add-ons with an international audience.

Note: Unsure if changing _name_ will interfere with AMO submission process or add-on validation.

shubheksha:
Fixes #148

### DIFF
--- a/context-menu-demo/manifest.json
+++ b/context-menu-demo/manifest.json
@@ -1,7 +1,8 @@
 {
 
   "manifest_version": 2,
-  "name": "context-menu-demo",
+  "name": "__MSG_extensionName__",
+  "description": "__MSG_extensionDescription__",
   "version": "1.0",
   "default_locale": "en",
 


kumar303:
Great, thanks for this!

PR #164, opened 2016-12-14
Add Omnibox API example

matthewjwein:


firefox-code-search/background.js
```diff
@@ -0,0 +1,83 @@
...
+});
+
+chrome.omnibox.onInputEntered.addListener((text, disposition) => {
+  let url = text;
+  if (!text.startsWith(SRC_URL)) {
+    url = `${SRC_URL}/search?q=${text}+-path%3Aobj-&redirect=false`
+  }
+  chrome.tabs.query({active: true, currentWindow: true}, tabs => {
+    chrome.tabs.update(tabs[0].id, {url});
+  });
```
andymckay: We've been trying to use `browser` and promises in the examples rather than callbacks.

firefox-code-search/background.js
```diff
@@ -0,0 +1,83 @@
+const SRC_URL = "http://searchfox.org/mozilla-central";
```
andymckay: Still sad this isn't https, but that's nothing you can do.

firefox-code-search/background.js
```diff
@@ -0,0 +1,83 @@
...
+    let contentType = response.headers.get("content-type");
+    if(contentType && contentType.indexOf("application/json") !== -1) {
+      response.json().then(function(json) {
+        let suggestions = [];
+        if (json.normal) {
+          if (json.normal["Textual Occurrences"]) {
+            let occurrences = json.normal["Textual Occurrences"].slice(0, 3);
+            occurrences.forEach(occurrence => {
+              suggestions.push({
+                content: `http://searchfox.org/mozilla-central/source/${occurrence.path}#${occurrence.lines[0].lno}`,
```
andymckay: Would it make sense to use SRC_URL here (and a few other places below)?

firefox-code-search/background.js
```diff
@@ -0,0 +1,83 @@
...
+  parts.forEach((part, i) => {
+    if (part.startsWith(PATH_PREFIX)) {
+      path = part.slice(PATH_PREFIX.length);
+    } else {
+      queryParts.push(part);
+    }
+  });
+
+  query = queryParts.join(' ');
+  return `${SRC_URL}/search?q=${query}&case=false&regexp=${isRegex}&path=${path}`;
```
andymckay: When testing this I get `isRegex is not defined` and I can't spot it anywhere else in here.

firefox-code-search/background.js
matthewjwein: It seems to work with https now so I updated it.

firefox-code-search/background.js
matthewjwein: Sounds good to me.

firefox-code-search/background.js
matthewjwein: Yeah, I need to remove that - I decided to leave support for regex out of the first version, but it wouldn't be hard to add.

firefox-code-search/background.js
matthewjwein: Yeah, it does, thanks.

matthewjwein:
I updated the patch based on the suggestions you made, and also added disposition handling to onInputEntered.

andymckay:
Thumbs up from me, works great and thanks for adding comment. Let's see what input @wbamberg has.

firefox-code-search/README.md
```diff
@@ -0,0 +1,13 @@
+# firefox-code-search
+
+## What it does
+
+This extension allows you to search the firefox codebase using the awesome bar.
```
wbamberg: firefox->Firefox

firefox-code-search/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+To test it out, type 'cs' into the awesome bar followed by a search string (e.g. `cs hello world`). The results will be shown as suggestions, and clicking on a suggestion will navigate to the file where the result was found.
+
+To search a specific file, use the "path:" prefix (e.g. `cs hello path:omnibox.js`)
+
+## What it shows
+
+How to use the omnibox API to create an extension which provides suggestions asynchronously.
```
wbamberg: It might be better to say something like "How to use the omnibox API to add custom suggestions to the awesome bar".

firefox-code-search/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
```
wbamberg: I think this isn't shipping until Firefox 52? So it would be good to have an `applications` key with `strict_min_version` of 52.0a1 (or something).

wbamberg:
Very nice example! I just had a few comments. Sorry to be slow getting them to you.

matthewjwein:
Thanks for the quick review! I made the changes you requested and also updated the URL match pattern to be more specific (from "<all_urls>" to "https://searchfox.org/").

wbamberg:
Thanks @matthewjwein !

PR #176, opened 2017-01-27
example of using topSites and chrome_url_override

andymckay:


andymckay:
maybe also cc @mixedpuppy and @matthewjwein, not sure if there's a better example we could do

top-sites/sites.js
```diff
@@ -0,0 +1,23 @@
...
+    var div = document.getElementById('site-list');
+
+    if (!sites.length) {
+      div.innerText = 'No sites returned from the topSites API.';
+      return;
+    }
+
+    let ul = document.createElement('ul');
+    ul.className = 'list-group';
+    for (let k in sites) {
```
wbamberg: Is there a reason we're using for-in rather than for-of https://hacks.mozilla.org/2015/04/es6-in-depth-iterators-and-the-for-of-loop/?

top-sites/manifest.json
```diff
@@ -0,0 +1,11 @@
...
+  "manifest_version": 2,
+  "name": "Override examples",
+  "version": "0.1",
+  "chrome_url_overrides": {
+      "newtab": "sites.html"
+  },
+  "permissions": [
+      "topSites"
+  ]
+}
```
wbamberg: I think this needs a strict_min_version: "54.0a1" or something like that.

top-sites/README.md
```diff
@@ -0,0 +1,9 @@
+# Top Sites
+
+## What it does
+
+This extension creates a new tab page that lists the topSites visited.
```
wbamberg: It might be clearer to say something like: "this extension replaces the built-in page shown when the user opens a new tab without specifying a page to load (for example, when the user presses Ctrl+T). The replacement page is populated with links taken from the topSites API."

wbamberg:
Nice example @andymckay ! I had just a couple of comments.

top-sites/sites.js
andymckay: There were some more features I added, then I figured the code was too complicated, removed it and forgot to fix it back up.

andymckay:
Fixed up

matthewjwein:
I just tested it out in nightly and it works great. Nice!

PR #180, opened 2017-02-04
Added an example using the identity API

wbamberg:


wbamberg:
@andymckay , @mixedpuppy, this is a simplified version of the identity example in https://github.com/mdn/webextensions-examples/pull/141. I've simplified it by removing the Facebook version and using Promise-based APIs.

Would you be able to review it? I'm especially interested in whether the OAuth flow is correct, and whether the use of promises here looks correct.

I think we might still want a separate example of the Facebook flow, but think it's clearer to have 2 examples here.

google-userinfo/background/main.js
```diff
@@ -0,0 +1,23 @@
+
```
andymckay: extra line

google-userinfo/background/main.js
```diff
@@ -0,0 +1,23 @@
...
+/*
+When the button's clicked:
+- get an access token using the identity API
+- use it to get the user's info
+- show a notification containing some of it
+*/
+browser.browserAction.onClicked.addListener(() => {
+  getAccessToken().then(getUserInfo)
+                  .then(notifyUser)
+                  .catch(logError);
```
andymckay: i thought the style was just to put .then unidented so lining up below getAccessToken, although we don't have much of a style guide.

google-userinfo/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+* **registering your add-on with Google as an OAuth2 client**.
+  * Visit https://console.developers.google.com/apis/credentials
+  * Click "Create credentials", and select "OAuth client ID"
+  * Select "Web application", and give it a name. The name is shown to the user to help them understand whether to authorize the add-on.
+  * Paste the redirect URL into the " Authorized redirect URIs" box.
+  * Click "Create"
+  * You'll see a popup containing a Client ID and a secret. Copy the client ID (you can ignore the secret).
+  * Paste this value into authorize.js in place of YOUR-CLIENT-ID.
+  * Reload the add-on.
```
andymckay: Because you need to alter authorise.js, we can't really sign this add-on.

Would it make sense to add in line about how this add-on has to be run through web-ext or about:debugging?

google-userinfo/background/main.js
```diff
@@ -0,0 +1,23 @@
...
+  console.error(`Error: ${error}`);
+}
+
+/*
+When the button's clicked:
+- get an access token using the identity API
+- use it to get the user's info
+- show a notification containing some of it
+*/
+browser.browserAction.onClicked.addListener(() => {
```
andymckay: If CLIENT_ID hasn't been changed, and the user hasn't gone through the setup step, would it make sense to error here?

google-userinfo/background/authorize.js
```diff
@@ -0,0 +1,74 @@
+const REDIRECT_URL = browser.identity.getRedirectURL();
+const CLIENT_ID = "YOUR-CLIENT-ID";
+const SCOPES = ["openid", "email", "profile"];
+const AUTH_URL =
```
mixedpuppy: nit indententation

google-userinfo/background/authorize.js
```diff
@@ -0,0 +1,74 @@
...
+function extractAccessToken(redirectUri) {
+  let m = redirectUri.match(/[#\?](.*)/);
+  if (!m || m.length < 1)
+    return null;
+  const params = new URLSearchParams(m[1].split("#")[0]);
+  return params.get("access_token");
+}
+
+/*
+Validate the token contained in redirectURL.
```
mixedpuppy: nit use jsdoc style comments (elsewhere also)

google-userinfo/background/authorize.js
```diff
@@ -0,0 +1,74 @@
...
+&response_type=token
+&redirect_uri=${encodeURIComponent(REDIRECT_URL)}
+&scope=${encodeURIComponent(SCOPES.join(' '))}`;
+const VALIDATION_BASE_URL="https://www.googleapis.com/oauth2/v3/tokeninfo";
+
+function extractAccessToken(redirectUri) {
+  let m = redirectUri.match(/[#\?](.*)/);
+  if (!m || m.length < 1)
+    return null;
+  const params = new URLSearchParams(m[1].split("#")[0]);
```
mixedpuppy: nit use let consistently

wbamberg: Do you mean `params` should be declared with `let`? I don't mean to go against the style guide (although it [doesn't seem to cover this case](https://wiki.mozilla.org/WebExtensions/Hacking#Code_Style)) but it seems more useful to use `const` where we can.

google-userinfo/background/authorize.js
wbamberg: Do you mean I should indent the pieces of AUTH_URL, like:

```
const AUTH_URL = `https://accounts.google.com/o/oauth2/auth
  ?client_id=${CLIENT_ID}
  &response_type=token
  &redirect_uri=${encodeURIComponent(REDIRECT_URL)}
  &scope=${encodeURIComponent(SCOPES.join(' '))}`;
```
? I didn't do this because then you get spaces in the URL. Sorry if I'm being slow-witted here though.

mixedpuppy: Right, ignore this one

google-userinfo/background/main.js
wbamberg: Like this:

```
browser.browserAction.onClicked.addListener(() => { 
   getAccessToken().then(getUserInfo)
   .then(notifyUser)
   .catch(logError);
```

It seems less readable to me, but if that's the rule, then OK. (I have no idea really how indentation should work with thens).

mixedpuppy: more I think the way you did it is ok, but for long lines generally something like:
```
call()
  .then()
  .catch()
```

andymckay: Making the indentation depend on the length of the method `getAccessToken()` is a real pain when you change the code later and use a different API then have to re-indent all the lines below it.

We ran into this so much in python that we just indent at a fixed limit like Shane suggests.

wbamberg: Yes, Shane's suggestion looks great to me :).

google-userinfo/background/authorize.js
mixedpuppy: I'm probably old school, and it is just a nit.  const tells me something about the variable, that it is being protected from change, for some reason.  In this case the variable is in a scope that cant be touched and the variable is dropped right away.  So to me, seeing const there makes my brain spin on "why".

google-userinfo/background/main.js
wbamberg: I didn't make this change, because you actually get a pretty good diagnostic error message from Google if the client ID doesn't match. Also, I feel like this is not a check you'd want to have in your production WebExtension (it's only really relevant because this is a generic example) and I wouldn't want people copying it into their production code.

wbamberg:
Thanks a lot for your comments :).

PR #181, opened 2017-02-04
add every event

andymckay:


andymckay:
screenshot 2017-02-03 21 58 59

andymckay:
This has proved pretty useful to me already. I doesn't actually show any API, but thought it would be useful for illustration purposes. I'm thinking I might just stick it on AMO though, its not like most of the other examples in this repository. What do you reckon @wbamberg?

wbamberg:
I tend to think this doesn't belong here. I think of these examples more about showing how to use the WE APIs, or as a starting point for people to create their own. and I'm not sure this fits either of those, although I could see it could be a useful tool for discovering which events are supported.

andymckay:
I'll move over to AMO then. Cheers.

Issue #182, opened 2017-02-09
bookmark-it bookmarks wrong tab


frostschutz:
When using more than one browser window, clicking the bookmark-it icon in Window X, instead of bookmarking tab X, instead it updates bookmark and bookmark-it icon in Window Y.

bookmark-it/background.js relies on `tabs.onUpdated` and `tabs.onActivated` events to update the icon on-off state and stores active tab/bookmark in `currentTab`, `currentBookmark` variables; for the `browserAction.onClicked` event it relies on those variables, but when using multiple windows those variables are not updated correctly.

I'm not sure how to solve this, other than bite the bullet, get rid of those variables, and re-run both `tabs.query({active: true, currentWindow: true})` and ` bookmarks.search()` in the browser action toggle function?

That might actually make the code shorter as `updateActiveTab()` and `toggleBookmark()` would be using the same logic then, so it could use the same function just with a toggle flag added. As a side effect it should also do the right thing if a bookmark was created/removed in the meantime while current bookmark-it still remembers a bookmark that no longer exists.

But I'm not sure if that is the correct solution.

wbamberg:
Thanks for filing, and for thinking about a fix! I'm afraid I didn't get a chance to think in detail about a solution. But [browserAction.onClicked passes the Tab for the current tab to its listener](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/browserAction/onClicked#addListener_syntax). Perhaps you could do something like:

* on click, get passed the current tab. Toggle its state (if your tab's icon is "star-filled", unbookmark it and switch icon, otherwise bookmark it and switch icon).

* when the active tab changes: get the active tab ID, then use it to search bookmarks, and use the result to set the icon.

?

I agree that it would be nice to get rid of those variables, they seem like asking for trouble.

PR #186, opened 2017-02-20
Rewrite of bookmark-it, fixing issue 182 and some other problems

wbamberg:
Here's a PR to fix https://github.com/mdn/webextensions-examples/issues/182. It's basically a rewrite, because it seems to me that storing `currentTab` and `currentBookmark` as global variables makes it easy for the add-on to get out of sync with reality. Also, the current add-on didn't handle changes to the bookmark store that happen outside the add-on itself (e.g. through the built-in bookmarking UI).

There's still a bug: bookmarks.search expects the `url` option to be in the `url` format which throws an error if the URL is like "about:debugging" [[1]](http://searchfox.org/mozilla-central/source/browser/components/extensions/schemas/bookmarks.json#258), [[2]](http://searchfox.org/mozilla-central/source/toolkit/components/extensions/Schemas.jsm#511). This seems like a WebExtensions bug, tbh, since it is possible to bookmark URLs like this, so it should be possible to search for those bookmarks too.

@kumar303 , do you think you could review this for me please?


bookmark-it/scripts/handle-button-click.js
```diff
@@ -0,0 +1,31 @@
...
+On button click:
+* find out if the tab's URL is bookmarked.
+*   if it isn't, bookmark it.
+*   if it is, remove the bookmark.
+*/
+function toggleBookmark(tabInfo) {
+
+  function updateBookmark(bookmarkArray) {
+    if (bookmarkArray.length === 0) {
+      browser.bookmarks.create({
```
kumar303: Since you're using a `catch` function down below, the returned promise should be returned to the chain:
````js
return browser.bookmarks.create({...});
````

File: bookmark-it/scripts/handle-button-click.js [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+        title: tabInfo.title,
+        url: tabInfo.url
+      });
+    } else {
+      browser.bookmarks.remove(bookmarkArray[0].id);
```
kumar303: same here:
````js
return browser.bookmarks.remove(bookmarkArray[0].id);
````

File: bookmark-it/scripts/update-icon.js [hunk 3]
```diff
@@ -0,0 +1,91 @@
...
+}
+
+
+/**
+If the array is empty, the page is not bookmarked.
+Otherwise, the page is bookmarked.
+*/
+function updateIcon(bookmarkArray) {
+  if (bookmarkArray.length === 0) {
+    browser.browserAction.setIcon({
```
kumar303: Similarly, this promise should be returned so your call below can catch errors:
````js
return browser.browserAction.setIcon({...});
````

File: bookmark-it/scripts/update-icon.js [hunk 4]
```diff
@@ -0,0 +1,91 @@
...
+      path: notBookmarked
+    });
+  } else {
+    browser.browserAction.setIcon({
```
kumar303: and this one too:
````js
return browser.browserAction.setIcon({...});
````

**Comment by wbamberg** [comment]
@kumar303 , sorry to take so long to get to this. I don't think {currentWindow: true} does help: the main problem in the old version was that it wasn't listening for window switching. But in the end I revised the fix to be much smaller in scope.

I've added:
* listening to windows.onFocusChanged so it works with multiple windows
* ignoring protocols that will raise an error in bookmarks.search, and logging the ignoring
* listening for bookmarks being created and removed

File: bookmark-it/background.js [hunk 5]
```diff
@@ -42,28 +36,46 @@ browser.browserAction.onClicked.addListener(toggleBookmark);
  */
 function updateActiveTab(tabs) {
 
+  function isSupportedProtocol(urlString) {
+    var supportedProtocols = ["https:", "http:", "ftp:", "file:"];
+    var url = document.createElement('a');
+    url.href = urlString;
+    return supportedProtocols.indexOf(url.protocol) != -1;;
```
kumar303: This line has an extra semi colon

wbamberg:
Thanks @kumar303 !

PR #192, opened 2017-03-04
Added sidebar example

wbamberg:
Here's a sidebar example.

It's a very simple page annotator. The sidebar just contains a box: when you mouseover the sidebar the box becomes contenteditable, and you can type your stuff. When you mouseout the contents of the box are stored, keyed to the currently active tab.

The sidebar then displays the content you saved for the currently active tab in the sidebar's window: when you switch tabs and stuff, the right content should be displayed.

I don't know if the mouseover/mouseout interaction model is a very good model. But I think the most 
useful thing about this example is that it shows how to keep the sidebar in sync with the active tab, which is not very obvious when you have multiple windows. I'm not sure if what I've done here is the right way, but it does seem to work.


annotate-page/sidebar/panel.css
```diff
@@ -0,0 +1,24 @@
+
```
andymckay: extra new line

annotate-page/sidebar/panel.js
```diff
@@ -0,0 +1,57 @@
...
+Make the content box editable as soon as the user mouses over the sidebar.
+*/
+window.addEventListener("mouseover", (e) => {
+  contentBox.setAttribute("contenteditable", true);
+});
+
+/*
+When the user mouses out, save the current contents of the box.
+*/
+window.addEventListener("mouseout", (e) => {
```
andymckay: nice

wbamberg:
Thanks @andymckay !

PR #198, opened 2017-03-21
Fix links in code of conduct

teoli2003:
The links had an extra space and weren't considered at links, this is now fixed.


wbamberg:
Thanks @teoli2003 !

PR #201, opened 2017-03-29
Changed URL to httpbin.org in the permissions

JonathanNakandala:
It looks like in a previous change the website used to demonstrate the
useragent rewriting was changed but the addon permission wasn't updated
so the demonstration didn't work.


andymckay:
Ah thanks, this would have been affected by https://bugzilla.mozilla.org/show_bug.cgi?id=1311815

PR #203, opened 2017-04-05
Add a keyboard shortcut to open the sidebar

wbamberg:
I thought it would be nice to show how to use _execute_sidebar_action in this example as well. @mixedpuppy, would you mind reviewing?


Natim:
In my browser Ctrl+Shift+Y opens the Library on the Download panel.

PR #204, opened 2017-04-07
an example using permissions

andymckay:
not too merge yet, can't seem to get permissions dialog actually showing up :(


andymckay:
it needs https://bugzilla.mozilla.org/show_bug.cgi?id=1352152 to land first

andymckay:
That bug is fixed, currently blocked on https://bugzilla.mozilla.org/show_bug.cgi?id=1336576

andymckay:
Actually that's not a blocker, it still works.

r? @wbamberg

permissions/README.md
```diff
@@ -0,0 +1,9 @@
+# Permissions Example
+
+## What it does
+
+An example using the permissions API to grant an optional permission.
```
wbamberg: -> "to grant and revoke optional permissions"

permissions/background.js
```diff
@@ -0,0 +1,5 @@
+browser.browserAction.onClicked.addListener(() => {
+  browser.tabs.create({
+    url: chrome.runtime.getURL("page.html")
```
wbamberg: chrome->browser

permissions/manifest.json
```diff
@@ -0,0 +1,24 @@
...
+    "default_icon": "icon.svg"
+  },
+
+  "permissions": [
+    "tabs"
+  ],
+  "optional_permissions": [
+    "history"
+  ]
+}
```
wbamberg: Since this API is new in Firefox 55, we should probably include an "applications" key here with strict_min_version

permissions/page.html
```diff
@@ -0,0 +1,24 @@
+<html>
```
wbamberg: There are some issues with the HTML here. I'm not sure which of them are nits.
* no DOCTYPE
* no `meta charset` declaration (this one gives an error in the console)
* no `<title>`
* `<div class="container">` is not closed (I think indentation would have made this more obvious)

permissions/page.html
```diff
@@ -0,0 +1,24 @@
...
+    </p>
+    <hr>
+    <ul>
+      <li>Tabs permission has been <b>granted</b> on install.</li>
+      <li>History permission is an optional permission that can be
+          <a href="#" class="permission" data-permission="history" data-action="grant">granted</a> or
+          <a href="#" class="permission" data-permission="history" data-action="revoke">revoked</a>.
+      </li>
+      <li>Cookies permission cannot be <a href="#" class="permission" data-permission="cookies" data-action="grant">granted</a>.</li>
+      <li>Foo permission cannot be <a href="#" class="permission" data-permission="foo" data-action="grant">granted</a>.</li>
```
wbamberg: These lines might be more easily understood if written like: "The `tabs` permission has been..."

It also might be informative to give reasons for the last two, like:

"The `cookies` permission cannot be granted, because it was not included in the optional_permissions manifest key."

"The `foo` permission cannot be granted, because it is not a valid permission name."

permissions/page.js
```diff
@@ -0,0 +1,41 @@
...
+function processChange(event) {
+  let permission = event.target.dataset.permission;
+  let result = document.getElementById('result');
+  let permissionMethod = (
+    event.target.dataset.action === 'grant' ?
+    browser.permissions.request : browser.permissions.remove
+  );
+
+  try {
+    permissionMethod({permissions: [permission]})
```
wbamberg: This is neat, but I wonder if it obscures the actual calls to the API, which are the main focus of this example. Just an opinion though :).

andymckay:
All good comments, updated.

wbamberg:
Thanks @andymckay !

PR #205, opened 2017-04-10
Add a proxy example

matthewjwein:


andymckay:
Are you missing a manifest.json file?

matthewjwein:
Yeah, I was, thanks

proxy/proxy_script.js
```diff
@@ -0,0 +1,18 @@
...
+  enabled: false,
+};
+
+// Required PAC function that will be called to determine
+// if a proxy should be used.
+function FindProxyForURL(url, host) {
+  if (settings.enabled) {
+    return "DIRECT 1.2.3.4:8080"; // Update to return proxy server rules (e.g. "PROXY 1.2.3.4:8080")
+  }
+  return "DIRECT 1.2.3.4:8080";
```
andymckay: This is the same as line 10.

proxy/background.js
```diff
@@ -0,0 +1,9 @@
+// Register the proxy script
+browser.proxy.registerProxyScript("proxy_script.js");
+
+// Called if there is an error.
+browser.proxy.onProxyError.addListener(error => {
+  console.log(error.message);
```
andymckay: Maybe you could change this something like ``` `Error in proxy: ${error.message}` ``` to make it clear to the user.

proxy/proxy_script.js
```diff
@@ -0,0 +1,18 @@
...
+}
+
+// Listen for messages from the background script.
+browser.runtime.onMessage.addListener(msg => {
+  settings.enabled = msg.enabled;
```
andymckay: So when does this get sent and the proxy enabled?

proxy/proxy_script.js
andymckay: So if I visit 1.2.3.4:8080 is something meant to happen? I'm a bit confused by this.

Could we use console.log in this proxy script to do something like "For domain X, using proxy at: ..." or something. Right now its a bit baffling to me what is meant to happen.

proxy/manifest.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "Proxy Boilerplate Code",
+  "manifest_version": 2,
+  "version": "1.0",
+  "background": {
+    "scripts": ["background.js"]
+  },
+  "permissions": [
+    "proxy"
+  ],
+  "description": "Provides boilerplate code to get started with the Proxy API.",
```
andymckay: Is it possible to make it do something, even just write a console.log?

proxy/manifest.json
```diff
@@ -0,0 +1,13 @@
...
+  "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/proxy"
+}
```
andymckay: Please add in an applications > strict_min_version that matches the version of Firefox where the API landed.

proxy/proxy_script.js
matthewjwein: This gets sent by the background script right when the extension is installed. The main purpose for this is to show that the proxy script supports runtime.onMessage.

proxy/proxy_script.js
matthewjwein: Nothing will happen. It's hard to really have anything useful without setting up a proxy server. 

It should really just be returning "DIRECT" here, not "DIRECT 1.2.3.4:8080", but I discovered a bug where "DIRECT" requires an argument to be passed for it to work - I've filed bug 1355198 to fix this.

proxy/background.js
matthewjwein: Sounds good, will do.

proxy/manifest.json
matthewjwein: The proxy script doesn't have access to console.log, but it can send messages to the background script which could log the message it receives. I'll add an example of this.

proxy/manifest.json
matthewjwein: Will do.

proxy/README.md
```diff
@@ -0,0 +1,9 @@
+# proxy-example
+
+## What it does
+
+This extension provides boilerplate code to get started writing an addon which uses the Proxy API.
+
```
wbamberg: I'd really like a bit more of a user-focused explanation of what this example does or why it might be useful. Even if it can't implement the whole use case, we could perhaps talk here about how it could be developed into something that addresses some use case or other.

andymckay:
@matthewjwein, I think we should probably close this one based on what @wbamberg landed.

matthewjwein:
Yep, I agree, thanks @wbamberg!

PR #206, opened 2017-04-11
Example of the webRequest.onAuthRequired API

wbamberg:
This add-on uses the webRequest.onAuthRequired to supply stored credentials to https://httpbin.org/basic-auth/user/passwd.

The options page to change the stored credentials is not really needed, and should perhaps be removed.

I'm not sure whether the way we keep track of requests is a good idea or a good implementation. But it seems like we do need a way to fail if the credentials are bad. Don't we?


wbamberg:
@mixedpuppy, would you be able to take a look at this?

wbamberg:
Thanks @mixedpuppy. I fixed the README.

PR #207, opened 2017-04-19
Add example with advanced clipboard manipulation

Rob--W:
The current clipboard example is too trivial.

I created an example that shows how to use the clipboard API from a background page.

(tested with Firefox 52 and 54)


wbamberg:
Very nice.

Issue #197, opened 2017-03-20
Cannot send/receive messages between background and content after reloading addon


mchubby:
Hi,

There is an error logged in Browser console **after reloading a temporary addon** and use of `browser.tabs.sendMessage` or `browser.runtime.sendMessage`
```
Error: Could not establish connection. Receiving end does not exist.         (unknown)
```

reproduced in FX 52 and 54.0a2 with steps:

- create a blank profile and start it, close all extra tabs
- Load an https URL
- open `about:debugging` in a new tab
- load a temporary add-on that uses message communication. For instance, cookie-bg-picker/manifest.json
- Click on reload button
- Switch to initial https tab
- Click on browser action ui and select a background

Expected:
background has changed

Actual:
nothing happens

- Open (Ctrl+Shift+J) browser console to see diag message

Notes:
seems to be the same when installing `cookie_bg_picker-1.0-an+fx.xpi` twice.

wbamberg:
Thanks for filing, and for nice clear steps to reproduce, and sorry to take so long to look at this. I've made a PR that should fix this: https://github.com/mdn/webextensions-examples/pull/212.

wbamberg:
Fixed: https://github.com/mdn/webextensions-examples/pull/212

PR #212, opened 2017-04-28
Use manifest key instead of executeScript to load content script

wbamberg:
This is a fix for https://github.com/mdn/webextensions-examples/issues/197.

I think the problem there is that the background script loads the content script using tabs.executeScript, but only does so on tabs.onUpdated. So if you install the add-on, then try to set a background for a page that was already open, the content script isn't yet loaded into that page. The popup then tries to send a message to the nonexistent script and you get an error.

The fix here is to load the content script using the content_scripts manifest key, so it's always there.

I've also changed the code that checks for cookies so it checks on activate as well as update.


chrisdavidmills:
Reviewed - this works really nicely. Thanks Will!

PR #218, opened 2017-05-11
Use webextensions env from eslint

freaktechnik:
eslint provides an env option for webextensions. Use that instead. It also seems the single rule left in the config is from react or similar - for sure not vanilla eslint.

### DIFF
--- a/eslint-example/.eslintrc.json
+++ b/eslint-example/.eslintrc.json
@@ -1,14 +1,8 @@
 {
-    "globals": {
-        "browser": true,
-        "chrome": true
-    },
-    "rules": {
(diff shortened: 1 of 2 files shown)


eslint-example/.eslintrc.json
```diff
@@ -1,16 +1,10 @@
...
-        "chrome": true
-    },
-    "rules": {
-        "no-set-state": "off"
-    },
     "env": {
         "browser": true,
-        "es6": true
+        "es6": true,
+        "webextensions": true
```
kumar303: ooh, nifty

PR #221, opened 2017-05-19
Add an example of using 'command' in contextMenus

wbamberg:
-> https://bugzilla.mozilla.org/show_bug.cgi?id=1318791


wbamberg:
@mixedpuppy, would you mind reviewing this update to context-menu-demo?

PR #225, opened 2017-05-28
Add proxy example

wbamberg:
This example implements a blocker using the proxy API. It's somewhat based on one of the examples here: https://calomel.org/proxy_auto_config.html.

It would be much simpler without all the mucking around with storage, but it seems like configuring the PAC script with user settings is the kind of thing people might want to know about, and it gives a good opportunity to demonstrate plenty of background<->PAC script messaging.


proxy-blocker/options/options.css
```diff
@@ -0,0 +1,15 @@
+
```
andymckay: leading new line

proxy-blocker/options/options.js
```diff
@@ -0,0 +1,25 @@
...
+
+function onError(e) {
+  console.error(e);
+}
+
+// On opening the options page, fetch stored settings and update the UI with them.
+browser.storage.local.get().then(updateUI, onError);
+
+// Whenever the contents of the textarea changes, save the new values
+blockedHostsTextArea.addEventListener("change", storeSettings);
```
andymckay: this will fire for every character? spamming browser.storage.local with every change seems like a bad example to set.

proxy-blocker/proxy/proxy-script.js
```diff
@@ -0,0 +1,21 @@
...
+browser.runtime.onMessage.addListener((message) => {
+  blockedHosts = message;
+});
+
+// required PAC function that will be called to determine
+// if a proxy should be used.
+function FindProxyForURL(url, host) {
+  if (blockedHosts.indexOf(host) != -1) {
+    browser.runtime.sendMessage(`Proxy-blocker: blocked ${url}`);
+    return deny;
```
andymckay: This is cool.

proxy-blocker/background/proxy-handler.js
```diff
@@ -0,0 +1,56 @@
...
+
+    })
+    .catch(()=> {
+      console.log("Error retrieving stored settings");
+    });
+}
+
+function handleMessage(message, sender) {
+  // only handle messages from the proxy script
+  if (sender.url !=  browser.extension.getURL("") + proxyScriptURL) {
```
andymckay: isn't that the same as browser.extension.getURL(proxyScriptURL)?

proxy-blocker/manifest.json
```diff
@@ -0,0 +1,24 @@
...
+    "scripts": [
+      "background/proxy-handler.js"
+    ]
+  },
+
+  "options_ui": {
+    "page": "options/options.html"
+  },
+
+  "permissions": ["proxy", "storage"]
```
andymckay: set a min_version in applications matching when the API landed?

proxy-blocker/options/options.js
wbamberg: No: https://developer.mozilla.org/en-US/docs/Web/Events/change#Description. I think it will fire:

* if the user exits the text box (by clicking outside it or tabbing)
* if the user closes the options page

So I think I'd like to keep this as it is..

proxy-blocker/background/proxy-handler.js
wbamberg: Er, yes.

Issue #167, opened 2016-12-16
Should beastify execute the content script outside the click handler?


wbamberg:
The tutorial on beastify got this change:

https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Your_second_WebExtension$compare?locale=en-US&to=1156264&from=1143411

...moving the executeScript call before the click handler, so we always run the script on every page.

I'm not sure if this is better or not. I think the reason I didn't do it before is that it seems wasteful to run the script on every page, since the user might only want to beastify a few pages. But as the comment points out, this also means that the script will be added twice if the button is clicked twice (I think, or is that not right?).

wbamberg:
Fixed by https://github.com/mdn/webextensions-examples/pull/240.

PR #227, opened 2017-06-02
Warning about injecting the content script multiple times

losnappas:
I think this example is something those new to addon developing will use as the base for their addon. Injecting the content script inside the listener isn't "normal" practice either(I think?), just happens to happen for this particular example.

source: happened to me *sob*

### DIFF
--- a/beastify/popup/choose_beast.js
+++ b/beastify/popup/choose_beast.js
@@ -24,6 +24,9 @@ If the click is on one of the beasts:
 If it's on a button wich contains class "clear":
   Reload the page.
   Close the popup. This is needed, as the content script malfunctions after page reloads.
+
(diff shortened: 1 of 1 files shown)


wbamberg:
@losnappas , I agree with you that we should do something about this. I filed this: https://github.com/mdn/webextensions-examples/issues/167 about the same issue.

Even if you move it out of the listener though, won't it still inject the script multiple times if you click the browser action multiple times?

There's an interesting bit of code here: https://github.com/mdn/webextensions-examples/blob/master/context-menu-copy-link-with-types/background.js#L23 that I've been thinking about adding beastify, to protect against running the script multiple times.

I'd like to get more input about good practice here.

wbamberg:
I've opened https://github.com/mdn/webextensions-examples/pull/240, which should hopefully make this problem go away.

PR #228, opened 2017-06-02
add in a theme switching example

andymckay:


theme-switcher/manifest.json
```diff
@@ -0,0 +1,29 @@
...
+  "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/theme-switcher",
+  "manifest_version": 2,
+  "name": "Theme Switcher",
+  "permissions": [
+    "management"
+  ],
+  "icons": {
+      "128": "star-half.svg"
+  },
+  "background": {
```
wbamberg: This key should be removed: "background.js" doesn't exist and this will give a warning on install.

theme-switcher/switcher.js
```diff
@@ -0,0 +1,21 @@
+var themeList = document.getElementById('theme-list');
+
+function enableTheme(e) {
+  browser.management.setEnabled(e.target.value, true);
+  e.preventDefault();
+  window.close();
+}
+
+browser.management.getAll().then((themes) => {
+  for (let theme of themes) {
```
wbamberg: This works, but... I thought it was weird to see `management.getAll()` only returning themes. It's incompatible with Chrome and is generally a surprising API.

Is it the intention with `getAll()` to never support other types, or is this just a first step? I saw [bug 1336908](https://bugzilla.mozilla.org/show_bug.cgi?id=1336908), but it didn't really elaborate on that.

I guess this is mostly a comment on the API rather than on this example, but maybe consider filtering the result:

    themes = themes.filter((e)=>{return e.type==="theme"})

That would make it more obvious and Chrome-compatible (maybe?) and would keep the code working in case we ever changed the behavior of getAll().

Or maybe it would be more confusing this way, since it would imply that we can return other types :).

andymckay: https://bugzilla.mozilla.org/show_bug.cgi?id=1289336 is about extending the API to return all extensions.

theme-switcher/manifest.json
andymckay: Oops, forgot about that, thanks.

theme-switcher/switcher.js
wbamberg: I guess you mean https://bugzilla.mozilla.org/show_bug.cgi?id=1282981? Since it looks likely that this will be implemented, can we filter the results, as above, so this example keeps working when that lands?

andymckay: Sorry, thank you finding the correct bug. Filtering sounds good.

PR #230, opened 2017-06-12
add in a dynamic theme example

andymckay:


dynamic-theme/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+## What it does
+
+Checks the users time and then flips between two different themes. A "moon theme" for after 8pm and before 8am and a "sun theme" the rest of the time.
+
+To flip themes you can change the system time.
+
+## What it shows
+
+How to use the dynamic theme API and how alarms.
```
wbamberg: How to use the dynamic theme and alarms APIs?

dynamic-theme/background.js
```diff
@@ -0,0 +1,50 @@
...
+  browser.theme.update({
+    images: {
+      headerURL: 'moon.jpg',
+    },
+    colors: {
+      accentcolor: '#000',
+      textcolor: '#fff',
+    }
+  });
+}
```
wbamberg: I think I'd prefer code that declared const variables with the theme data in them, like:

    const day = {
      images: {
        headerURL: 'sun.jpg',
      },
      ...
    }

...then instead of `setSun()` and `setMoon()`, have `setTheme(theme)` called from `checkTime()` with the appropriate variable.

But this is just a matter of opinion.

dynamic-theme/background.js
```diff
@@ -0,0 +1,50 @@
...
+  } else {
+    setMoon();
+  }
+}
+
+// On start up, check the time to see what theme to show.
+checkTime();
+
+// Set up an alarm to check this regularly.
+browser.alarms.create('checkTime', {periodInMinutes: 5});
```
wbamberg: The first argument here isn't a function, it's just a name for the alarm (https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/alarms/create#Parameters).

So I think you need something like:

    browser.alarms.onAlarm.addListener(checkTime);

dynamic-theme/manifest.json
```diff
@@ -0,0 +1,14 @@
...
+  "name": "Dynamic theme example",
+  "permissions": [
+    "alarms",
+    "theme"
+  ],
+  "background": {
+    "scripts": ["background.js"]
+  },
+  "version": "1.0"
+}
```
wbamberg: We should add `strict_min_version`, since this is a new API.

wbamberg:
Thanks for making those changes. It looks and works fine.

andymckay:
Sorry, some stupid mistakes there, I was rushing.

PR #232, opened 2017-06-18
proxyAPI sample that uses proxy servers for proxying Firefox connections

ericjung:
The other proxy sample, proxy-blocker, demonstrates using a non-existent proxy server to "null route" selected requests by domain. That is atypical usage of proxy servers in web browsers. Typical usage is to route requests through a proxy server in order to use a different IP address for the request. This example demonstrates that use as well as authentication with proxy servers.


proxy-server/proxy/proxy-script.js
```diff
@@ -0,0 +1,19 @@
...
+});
+
+// required PAC function that will be called to determine
+// if a proxy should be used.
+function FindProxyForURL(url, host) {
+	browser.runtime.sendMessage("In FindProxyForURL()");
+	if (proxyInfo.proxyServerAddress && proxyInfo.proxyServerPort && proxyInfo.proxyServerType)
+  	return proxyInfo.proxyServerType + " " + proxyInfo.proxyServerAddress + ":" + proxyInfo.proxyServerPort;
+	else
+		return "DIRECT 123"; // After https://bugzilla.mozilla.org/show_bug.cgi?id=1355198 is fixed, change this to "DIRECT"
```
andymckay: some { and } might make this if statement a little clearer.

proxy-server/README.md
```diff
@@ -0,0 +1,30 @@
...
+* install it
+* open options and specify proxy address and port. username/password optional and they do not work with SOCKS proxy servers
+* check your IP address in firefox. it should be the IP address of the proxy server.
+
+## What it shows
+
+* How to implement a simple PAC script, and register it using the proxy API.
+* How to exchange messages between a PAC script and a background script.
+* How to store proxy info using the storage API
+* How to authenticate with HTTP and HTTPS/SSL proxy servers (but not SOCKS proxy servers)
```
andymckay: I was going to try this, but I don't have a proxy server, is there a simple example you could recommend? A javascript or python one liner that pulls in a repo would be a good example.

proxy-server/README.md
```diff
@@ -0,0 +1,30 @@
...
+If the proxy server requires authentication (by sending HTTP status code 407), then authentication info is sent to the proxy server. Authentication does not work for SOCKS servers. Note: after the following bugs are fixed, authentication
+info can be specified in the FindProxyForURL() function without need of browser.webRequest.onAuthRequired():
+
+* https://bugzilla.mozilla.org/show_bug.cgi?id=1359693
+* https://bugzilla.mozilla.org/show_bug.cgi?id=1359543
+* https://bugzilla.mozilla.org/show_bug.cgi?id=1360404
+
+
+To try it out:
+* install it
```
andymckay: Maybe something like "web-ext run" will be clearer

proxy-server/README.md
```diff
@@ -0,0 +1,30 @@
...
+* open options and specify proxy address and port. username/password optional and they do not work with SOCKS proxy servers
+* check your IP address in firefox. it should be the IP address of the proxy server.
```
andymckay: Capital "I" in "it", capital "f" on "Firefox" please.

proxy-server/background/proxy-handler.js
```diff
@@ -0,0 +1,69 @@
...
+  return browser.storage.local.get('authCredentials');
+}
+
+browser.runtime.onMessage.addListener(handleMessage);
+
+// After https://bugzilla.mozilla.org/show_bug.cgi?id=1359693 is fixed, onAuthRequired() not needed
+// You can specify username/password in FindProxyForURL() 
+browser.webRequest.onAuthRequired.addListener(provideCredentialsASync,
+                                          {urls: ["<all_urls>"]},
+                                          ["blocking"]);
```
andymckay: We should probably file an issue to clean this up once those bugs have landed :)

proxy-server/options/options.js
```diff
@@ -0,0 +1,56 @@
+const proxyServerAddressInput = document.querySelector("#proxyServerAddressInput");
+const proxyServerPortInput = document.querySelector("#proxyServerPortInput");
+const proxyServerUsernameInput = document.querySelector("#proxyServerUsernameInput");
+const proxyServerPasswordInput = document.querySelector("#proxyServerPasswordInput");
+const proxyServerTypeInput = document.querySelector('input[name=proxyServerTypeInput]:checked');
```
andymckay: nit: all the other values used ", except this one which is '

proxy-server/options/options.js
```diff
@@ -0,0 +1,56 @@
...
+	// authCrendentials object is handed directly to browser.webRequest.onAuthRequired()
+	// so it's variable name format varies from our other variables in order to fulfill the API specifcation
+	let authCredentials = {
+		username: proxyServerUsernameInput.value,
+		password: proxyServerPasswordInput.value,
+	};
+
+	console.log("Storing settings");
+  browser.storage.local.set({
+    proxyServerInfo, authCredentials
```
andymckay: nit: the indentation seems inconsistent, most of the other examples have gone for 2 spaces. This seems to jump around between 2 spaces and 8. Also that's what you get when python developers review things.

proxy-server/README.md
```diff
@@ -0,0 +1,30 @@
+# proxy-addon
+
+## What it does
+
+This add-on registers a [PAC script](https://developer.mozilla.org/en-US/docs/Web/HTTP/Proxy_servers_and_tunneling/Proxy_Auto-Configuration_%28PAC%29_file) using the proxy API.
+
+The PAC script used to load all requests through a proxy server. If no proxy server address or proxy server
```
wbamberg: "script used" -> "script is used"

proxy-server/background/proxy-handler.js
```diff
@@ -0,0 +1,69 @@
...
+  console.error(`Proxy error: ${error.message}`);
+});
+
+// Initialize the proxy
+function handleInit(message) {
+  // update the proxy whenever stored settings change
+  browser.storage.onChanged.addListener((newSettings) => {
+    if (newSettings.proxyServerInfo) {
+      console.log("Sending updated proxyServerInfo to proxy script");
+      browser.runtime.sendMessage(newSettings.proxyServerInfo.newValue, {toProxyScript: true});
```
wbamberg: Since storage is updated whenever any of the options pages textareas change, the proxy script will temporarily get inconsistent and potentially incorrect information if the user updates, say, host and port together. I don't know how much this matters really, but it's technically a bug.

One way to resolve it would be to have an explicit "save" button in the options page, that updates storage with all the new settings together.

proxy-server/background/proxy-handler.js
```diff
@@ -0,0 +1,69 @@
...
+
+  // get the current settings, then...
+  browser.storage.local.get()
+    .then((storedSettings) => {
+      // if there are stored settings, update the proxy with them...
+      if (storedSettings.proxyServerInfo) {
+        logSettings(storedSettings);
+        console.log("Sending updated proxyServerInfo to proxy script");
+        browser.runtime.sendMessage(storedSettings.proxyServerInfo, {toProxyScript: true});
+      }
```
wbamberg: I think it would be helpful to log something in the else case, like "Proxy settings have not been initialized" or "No proxy server defined: visit the options page to define a proxy server". Otherwise, AIUI, the default behavior will be to do nothing, with no output.

proxy-server/background/proxy-handler.js
```diff
@@ -0,0 +1,69 @@
...
+    })
+    .catch(()=> {
+      console.log("Error retrieving stored settings");
```
wbamberg: console.error maybe?

proxy-server/background/proxy-handler.js
```diff
@@ -0,0 +1,69 @@
...
+    });
+}
+
+function logSettings(settings) {
+  console.log("Got settings:");
+  console.log(`${settings.proxyServerInfo.proxyServerAddress}`);
+  console.log(`${settings.proxyServerInfo.proxyServerPort}`);
+  console.log(`${settings.proxyServerInfo.proxyServerType}`);
+  console.log(`${settings.authCredentials.username}`);
+  console.log(`${settings.authCredentials.paΩssword}`);
```
wbamberg: paΩssword -> password

proxy-server/background/proxy-handler.js
```diff
@@ -0,0 +1,69 @@
...
+
+  if (message === "init") {
+    handleInit(message);
+  } else {
+    // after the init message the only other messages are status messages
+    console.log(message);
+  }
+}
+
+function provideCredentialsASync(details) {
```
wbamberg: "provideCredentialsASync" -> "provideCredentialsAsync"

proxy-server/background/proxy-handler.js
```diff
@@ -0,0 +1,69 @@
...
+  return browser.storage.local.get('authCredentials');
+}
+
+browser.runtime.onMessage.addListener(handleMessage);
+
+// After https://bugzilla.mozilla.org/show_bug.cgi?id=1359693 is fixed, onAuthRequired() not needed
+// You can specify username/password in FindProxyForURL() 
+browser.webRequest.onAuthRequired.addListener(provideCredentialsASync,
```
wbamberg: What happens if the username and password are bad? Will the server just give up after a few tries?

proxy-server/manifest.json
```diff
@@ -0,0 +1,32 @@
+{
+
+  "manifest_version": 2,
+  "name": "Proxy Server",
+  "description": "Uses the proxy API to send all requests through a proxy server, optionally with authentication. Stores settings using the storage API.",
+  "version": "1.0",
+
+  "icons": {
+    "48": "icons/block.svg",
+    "96": "icons/block.svg"
```
wbamberg: Since this isn't blocking, we should use a different icon.

proxy-server/options/options.html
```diff
@@ -0,0 +1,29 @@
...
+      <input id="proxyServerPortInput"/><br>
+      <label>Proxy server username (optional)</label>
+      <input id="proxyServerUsernameInput"/><br>
+      <label>Proxy server password (optional)</label>
+      <input id="proxyServerPasswordInput"/><br>
+      <label>Proxy server type</label><br>
+      <input name="proxyServerTypeInput" type="radio" value="HTTP" checked>HTTP
+      <input name="proxyServerTypeInput" type="radio" value="HTTPS">HTTPS/SSL
+      <input name="proxyServerTypeInput" type="radio" value="SOCKS">SOCKS5
+    </p>
```
wbamberg: This looks pretty rough, because the boxes are all aligned differently. It would look better if the boxes came before the labels: http://imgur.com/a/lxwa3

proxy-server/proxy/proxy-script.js
wbamberg: This is fixed in 56, so this line now throws an error. That also means strict_min_version in manifest.json should be updated to "56.0a1".

proxy-server/options/options.js
```diff
@@ -0,0 +1,56 @@
...
+
+// Store the currently selected settings using browser.storage.local.
+function storeSettings() {
+  let proxyServerInfo = {
+		proxyServerAddress: proxyServerAddressInput.value,
+		proxyServerPort: proxyServerPortInput.value,
+		proxyServerType: proxyServerTypeInput.value
+	};
+
+	// authCrendentials object is handed directly to browser.webRequest.onAuthRequired()
```
wbamberg: "authCrendentials" -> "authCredentials"

proxy-server/options/options.js
```diff
@@ -0,0 +1,56 @@
...
+	// so it's variable name format varies from our other variables in order to fulfill the API specifcation
```
wbamberg: "it's" -> "its"

proxy-server/README.md
ericjung: "install it" is just copied/pasted from the proxy-blocker webextensions-example. However, I'll add some basic "web-ext run" info.

proxy-server/background/proxy-handler.js
ericjung: This is intended as a demonstration of the API. I do not think every possible corner case of potential pitfalls needs to be accounted for. It can convolute the example. I defer to you and Andy, however. Please let me know if you want a "save" button.

proxy-server/background/proxy-handler.js
ericjung: Does that get filed in github or bugzilla?

proxy-server/manifest.json
ericjung: Can you please suggest where to get these? I found something relevant: https://thenounproject.com/search/?q=proxy&i=831371 but then I noticed the license.

proxy-server/background/proxy-handler.js
ericjung: Nevermind. I actually agree :)

proxy-server/background/proxy-handler.js
andymckay: File it here since its changing code is this repo.

ericjung:
> this is failing the eslint test
Those failures are expected and should not be corrected. One failure says the FindProxyForURL() functions is not used. It is used, but not in the module in which it is defined. It is called by the WebExtension proxyAPI. The API requires it to exist. The other failure says the 2nd argument in FindProxyForURL() is not used. That's correct. I do not want to remove it from the function definition because this is a demo/example -- people are going to copy/paste parts of it, including FindProxyForURL(). Many real-world addons are going to need that 2nd argument; this one happens not to need it because it's very simple.

proxy-server/background/proxy-handler.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+});
+
+// get the current settings, then...
+browser.storage.local.get()
+  .then((proxyServerInfoAndAuth) => {
+    // attempt to register the proxy script
+    registerProxyScript(proxyServerInfoAndAuth);
+  })
+  .catch(()=> {
```
wbamberg: The old code distinguished between an error getting the settings, and successfully getting empty settings. This enabled it to log a helpful message on install:

```
  // get the current settings, then...
  browser.storage.local.get()
    .then((storedSettings) => {
      // if there are stored settings, update the proxy with them...
...
      }
      else {
        console.log("No proxy server defined: visit the options page to define a proxy server");
      }
    })
    .catch(()=> {
      console.error("Error retrieving stored settings");
    });
```
The new code doesn't do this, so you just see a generic storage error on startup.

proxy-server/background/proxy-handler.js
```diff
@@ -0,0 +1,114 @@
...
+function registerProxyScript(proxyServerInfoAndAuth) {
+  logSettings(proxyServerInfoAndAuth);
+  if (!proxyScriptLoaded) {
+    if (validateProxyServerInfo(proxyServerInfoAndAuth.proxyServerInfo)) {
+      console.log("Registering proxy script");
+
+      // Register the proxy script. Path is relative to manifest.json
+      browser.proxy.register(proxyScriptURL);
+      proxyScriptLoaded = true;
+      sendSettingsToProxyScript(proxyServerInfoAndAuth.proxyServerInfo);
```
wbamberg: I would have thought this would race: you might send the message before the proxy script has run and added its listener. But it seems to work...

proxy-server/background/proxy-handler.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+}
+
+function unregisterProxyScript() {
+  if (proxyScriptLoaded) {
+    console.log("Unegistering proxy script");
```
wbamberg: Typo

proxy-server/proxy/proxy-script.js
```diff
@@ -0,0 +1,24 @@
+var proxyInfo = {};
+
+// tell the background script that we are ready
+//browser.runtime.sendMessage("init");
```
wbamberg: Please remove this commented-out function.

wbamberg:
@ericjung , will you have the time to finish this example?

PR #234, opened 2017-06-19
move from url_overrides to settings_overrides

andymckay:
fixes change from https://bugzilla.mozilla.org/show_bug.cgi?id=1341458

### DIFF
--- a/top-sites/manifest.json
+++ b/top-sites/manifest.json
@@ -5,12 +5,16 @@
   "chrome_url_overrides": {
       "newtab": "sites.html"
   },
+  "chrome_settings_overrides": {
+      "homepage": "sites.html"
+  },
   "permissions": [
       "topSites"
   ],
   "applications": {
       "gecko": {
-          "strict_min_version": "54.0a1"
+          "strict_min_version": "55.0a1",
+          "id": "foo@bar.com"
       }
   }
 }


wbamberg:
@andymckay I'm not sure why this is called "fix-top-sites". The top-sites example doesn't set the homepage, it sets the new tab page (i.e. the page you get when you click Ctrl+T), and the behavior of `chrome_url_settings.newtab` is not affected by [bug 1341458](https://bugzilla.mozilla.org/show_bug.cgi?id=1341458).

So this PR adds a new feature - setting the homepage - rather than fixing an existing one. Which is fine, but I think it would be worth updating the [README](https://github.com/mdn/webextensions-examples/blob/master/top-sites/README.md) for.

Having said that, I tried it out, and while it works fine when you click the little house icon, it does not work when you type `about:home` into the address bar. Is that the intention? In Chrome, with this add-on, typing the equivalent `chrome://newtab` does display the custom version.

andymckay:
You know what, I had fiddled with this add-on locally and adding in the home page to test its behaviour. So I was fixing it locally and adding it in remotely. Doh.

Is it worth adding in as an example?

As for about:home, I didn't know about that, we should probably file a bug.

wbamberg:
I think it's fine to add it as an example (it might be good actually, since the `chrome_settings_overrides`/`chrome_url_overrides` business is pretty confusing).

Maybe it would be better to use a different page, so people can see a difference between newtab page and homepage? But then I suppose it would be weird in the context of this add-on, which is all about the topsites API.

Maybe... the homepage could stay the same, showing a list from topsites, and the newtab page could load a random site from topsites? 

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1375324.

andymckay:
One thing we should do is encourage an add-on to do one thing and keep the example small. I'll try to come up with a more relevant home page example.

andymckay:
Thanks for the bug.

PR #237, opened 2017-06-27
Hellosct1 webext

hellosct1:
New demo with devtools and the webextensions


devtools-panels/background_scripts/background.js
```diff
@@ -0,0 +1,19 @@
+
+/**
+When we receive the message, execute the given script in the given
+tab.
+*/
+function handleMessage(request, sender, sendResponse) {
+
```
wbamberg: I think it would be good for this to check where the message came from, especially as it's blindly executing some code in the tab. I think you should be able to do this using some code like:

```
if (sender.url != browser.runtime.getURL("/devtools/panel/panel.html")) {
  return;
}
```

devtools-panels/background_scripts/background.js
```diff
@@ -0,0 +1,19 @@
...
+/**
+When we receive the message, execute the given script in the given
+tab.
+*/
+function handleMessage(request, sender, sendResponse) {
```
wbamberg: `sendResponse` is not used and can be omitted (`sender` also, but see the next comment).

devtools-panels/devtools/panel/devtools-panel.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+Handle the result of evaluating the script.
+If there was an error, call handleError.
+*/
+function handleResult(result) {
+  console.log('Function handleResult');
+  console.log(result);
+  if (result[0] !== undefined) {
+    console.log(`jQuery: ${result[0]}`);
```
wbamberg: I think it's fragile to have this jQuery-specific handling in here. This function is called also to check the result of the other scripts: if one of them happens to not evaluate to undefined, then this will log misleading stuff.

It would probably be better to have a handler specifically for jQuery, then another generic one that only checks for errors.

devtools-panels/devtools/panel/devtools-panel.js
```diff
@@ -0,0 +1,71 @@
...
+document.getElementById("jquery").addEventListener("click", () => {
+  browser.devtools.inspectedWindow.eval(checkjQuery)
+    .then(handleResult);
+});
+
+/**
+When the user clicks each of the first three buttons,
+evaluate the corresponding script.
+*/
+const evalButton = document.querySelector("#reddinate");
```
wbamberg: this line is unused

devtools-panels/devtools/panel/devtools-panel.js
```diff
@@ -0,0 +1,71 @@
...
+const evalString = "$0.style.backgroundColor = 'red'";
+document.getElementById("background").addEventListener("click", () => {
+  browser.devtools.inspectedWindow.eval(evalString)
+    .then(handleResult);
+});
+
+const inspectString = "inspect(document.querySelector('h1'))";
+document.getElementById("h1").addEventListener("click", function(){
```
wbamberg: This does not use arrow notation, which is inconsistent with the others.

devtools-panels/devtools/panel/devtools-panel.js
```diff
@@ -0,0 +1,71 @@
...
+send a message to the background script.
+*/
+const scriptToAttach = "document.body.innerHTML = 'Hi from the devtools';";
+document.getElementById("message").addEventListener("click", () => {
+  browser.runtime.sendMessage({
+    tabId: browser.devtools.inspectedWindow.tabId,
+    script: scriptToAttach
+  });
+  
+  newPanel.sendMessage({
```
wbamberg: this generates an error, `newPanel` is not defined. I think you should just remove it.

devtools-panels/devtools/panel/panel.html
```diff
@@ -0,0 +1,14 @@
...
+<html>
+  <head>
+    <meta charset="utf-8">
+  </head>
+  <body>
+     <button id="h1">Inspect h1</button>
+     <button id="background">Reddinate inspected element</button>
+     <button id="jquery">Check for jquery</button>
+     <button id="message">Inject content script</button>    
+    <script src="devtools-panel.js"></script>
```
wbamberg: I also gave these longer, more descriptive labels. I hope that's OK with you.

hellosct1:
hi,
I made the various corrections proposed
I changed the image and I realized myself the new image (which is free)
Thnk's

devtools-panels/devtools/panel/devtools-panel.js
wbamberg: I don't think this comment was addressed in the last round?

devtools-panels/devtools/panel/devtools-panel.js
```diff
@@ -0,0 +1,65 @@
...
+Handle errors from the injected script.
+Errors may come from evaluating the JavaScript itself
+or from the devtools framework.
+See https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/devtools.inspectedWindow/eval#Return_value
+*/
+function handleError(error) {
+  if (error.isError) {
+    console.log('Devtools error: ${error.code}');
+  } else {
+    console.log('JavaScript error: ${error.value}');
```
wbamberg: Changing from `` to '' here and at line 9 introduces an error, please see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

hellosct1:
I just made the new corrections

devtools-panels/devtools/panel/devtools-panel.js
wbamberg: This issue still isn't addressed. Perhaps you could rename this function to handlejQueryTest, or something, then for the others, use an anonymous function, like:

  ```
browser.devtools.inspectedWindow.eval(evalString)
    .then(result => {
      if (result[1]) {
        handleError(result[1]);
      }
    });
```

hellosct1:
I think this time, it corresponds to what you expect. Thank's

hellosct1:
I made the corrections
I hope this is what you expect because otherwise I do not speak

PR #240, opened 2017-07-07
Improve content script injection

wbamberg:
This is an attempt to fix a couple of problems in the way beastify injects its content script.

Problem 1: before, the background script called executeScript then immediately tried to send the content script a message, before the content script had set up its listener. So you'd see "Error: Could not establish connection. Receiving end does not exist.".

Problem 2: the background script would re-inject the content script every time the user clicks the popup button. The content script did at least remove its listener after beastifying, but even so, this doesn't seem like very good practice.

In this PR, I've stolen this pattern: https://github.com/mdn/webextensions-examples/blob/master/context-menu-copy-link-with-types/background.js#L23-L33 from @Rob--W. This injects a content script as code, that just checks whether the `beastify()` function is defined in the page. If it's not, we inject the real content script, otherwise we do nothing. I've also updated the background script to send the message only after `executeScript`'s promise is fulfilled. I also updated the content script to not remove its listener, since that shouldn't be needed any more.

I hope this is more solid now. I think since this example is quite prominent, people will probably copy it, so it matters to make it good enough.


Rob--W:
I suggest to use a guard variable, because it is not prone to race conditions (tab navigating between the two injections) and involves less round-trips. For example:

```javascript
(function() {
    if (window.hasRun) return;
    window.hasRun = true;
    // Rest of logic 
})();
```

Noteworthy things:

- It uses a closure to not leak any variables or state to the global namespace.
- It uses one global variable to track state (toggle / untoggle).
-

wbamberg:
But in that case won't we inject the script multiple times, if we close and then reopen the popup?

Rob--W:
Yes it will, but the logic is executed only once because of the guard variable.

wbamberg:
Oh sorry I misunderstood you. I've updated this now, I think.

wbamberg:
@andymckay , can you review this? Unless you think @Rob--W would be a better choice.

andymckay:
I trust @Rob--W more than me, so if he's got the time...

beastify/content_scripts/beastify.js
```diff
@@ -1,37 +1,48 @@
...
-    document.body.firstChild.remove();
+  /*
+  Given a URL to a beast image, create and style an IMG node pointing to
+  that image, then insert the node into the document.
+  */
+  function insertBeast(beastURL) {
+    var beastImage = document.createElement("img");
+    beastImage.setAttribute("src", beastURL);
+    beastImage.setAttribute("style", "width: 100vw");
+    beastImage.setAttribute("style", "height: 100vh");
```
Rob--W: These two lines don't work, the last line overwrites the previous one.

Try

```javascript
beastImage.style.width = "100vw";
beastImage.style.height = "100vh";
```

or

```javascript
beastImage.style.cssText = "width: 100vw; height: 100vh;";
```

beastify/popup/choose_beast.js
```diff
@@ -16,33 +16,39 @@ function beastNameToURL(beastName) {
...
       browser.tabs.sendMessage(tabs[0].id, {beastURL: chosenBeastURL});
     });
   }
+
+  function reportError(error) {
+    console.error(`Could not inject content script: ${error}`);
+  }
+
+  if (e.target.classList.contains("beast")) {
+    browser.tabs.executeScript({
```
Rob--W: In this specific example, the current window and tab are not expected to change because the script is running from an extension popup. However, this coding pattern should not be used in a background page, because there is no guarantee that the "current tab" at the time of invoking `browser.tabs.executeScript` matches the "active tab in the current window" (lines 33-35) at the time of sending the message. This situation occurs when the user switches tabs or windows between the two calls.

To have this guarantee, call `browser.tabs.query` in advance and use the retrieved `tabId` for both `sendMessage` and `executeScript`.

You can either implement this suggestion, or add a note in the comments to warn the developer about this.

losnappas:
In my opinion, the problem still exists in this part
`if (e.target.classList.contains("beast")) {
    browser.tabs.executeScript({
      file: "/content_scripts/beastify.js"
    }).then(messageBeastify)
      .catch(reportError);
  }
  else if (e.target.classList.contains("clear")) {
    browser.tabs.reload();
    window.close();
    return;
}`
Considering this is "Your Seconds Web Extension", it's bizarre to inject the content script multiple times within a session because the page reloads. Your average addon does not reload the page and that way un-inject the content script. Adding a comment like "reloading the page un-injects the content script" would be a nice start. People will still use this as their base for their first addon and just remove the tabs.reload(), thus ending up injecting the content script multiple times again.

wbamberg:
@Rob--W , please take another look.

> Try
> 
> beastImage.style.width = "100vw";
> beastImage.style.height = "100vh";

In the end I thought just setting height is likely to work better.

wbamberg:
@losnappas, I'm not sure what you are asking for here. Do you think we should remove the code that reloads the page? Or are you suggesting that we should avoid injecting another content script if the user clicks a different button while the popup is open (which seems to be unrelated to the page reload question).

losnappas:
@wbamberg 

> Do you think we should remove the code that reloads the page?

yeah!! I think it makes the extension a bit odd. Usually your addons don't reload pages and then inject the content script again.

I know how it works now, but to a newcomer, it is not obvious logic. One might end up injecting their content scripts multiple times without reloading. At least it has the guard variable to cover for that now, so it might be fine.

Rob--W:
The extra changes looks good to me. I agree with @losnappas that reloading the page is a bit odd/user-unfriendly, but to not make the example overly complex, I think that it is fine to drop a comment why you reload the page, in a way that it is obvious that there are also negative effects (e.g. data loss)..

Or if you think that you can "undo" the changes without too much complexity, then by all means go for it.

wbamberg:
Looking at the code now, I wonder why we don't just inject the script once, when the popup loads:

```
browser.tabs.executeScript({
  file: "/content_scripts/beastify.js"
}).then(() => {
  document.addEventListener("click", (e) => {

    function reportError(error) {
      console.error(`Could not beastify: ${error}`);
    }

    function beastify(tabs) {
      var chosenBeast = e.target.textContent;
      var url = beastNameToURL(chosenBeast);
      browser.tabs.sendMessage(tabs[0].id, {beastURL: url});
    }

    if (e.target.classList.contains("beast")) {
      browser.tabs.query({active: true, currentWindow: true})
        .then(beastify)
        .catch(reportError);
    }
    else if (e.target.classList.contains("clear")) {
      browser.tabs.reload();
      window.close();
      return;
    }
  });
  
});
```

I think we still need the `window.close()` and the multiple-inclusion guard in the content script (in case the user reopens the popup in the same tab) but this looks more logical than injecting the script every time the user clicks a button inside the popup. Is this closer to what you were looking for, @losnappas ?

losnappas:
How about doing a ` document.body.cloneNode(true) ` on the first content script injection, then retrieving that on consecutive reset button clicks instead of reloading tab? As @Rob--W suggested

```javascript
window.cloned = document.body.cloneNode(true);
  /*
  beastify():
  * removes every node in the document.body,
  * then inserts the chosen beast
  */
  function beastify(request, sender, sendResponse) {
    removeEverything();
    if (request.beastURL.indexOf("Reset")!==-1)
      document.body.append(window.cloned); //bug: 2 body elements; couldn't figure it out
    else{
      insertBeast(request.beastURL);
    }
  }
```
now we can remove tabs.reload and window.close
```javascript
function beastify(tabs) {
    browser.tabs.executeScript(tabs[0].id, {
      file: "/content_scripts/beastify.js"
    }).then(() => {
      var chosenBeast = e.target.textContent;
      var url = chosenBeast;
      if (chosenBeast.indexOf("Reset")===-1)
        url = beastNameToURL(chosenBeast);

      browser.tabs.sendMessage(tabs[0].id, {beastURL: url});
    });
  }

  browser.tabs.query({active: true, currentWindow: true})
    .then(beastify)
    .catch(reportError);
```
@wbamberg is this too complex? I'd say this method would be a more typical addon structure.

Rob--W:
> How about doing a `document.body.cloneNode(true)` on the first content script injection, then retrieving that on consecutive reset button clicks instead of reloading tab?

That still has side effects, e.g. child frames are reloaded.

If you want to be able to restore the page, use CSS to hide all of the body's immediate children, except for the beastify images.

E.g. add `beastImage.className = "beastify-image";` and use tabs.insertCSS/removeCSS to toggle it.

```css
body > *:not(img.beastify-image) {
    display: none;
}
```

wbamberg:
> If you want to be able to restore the page, use CSS to hide all of the body's immediate children, except for the beastify images.

That's a clever idea. I've had a go at doing this, and at only injecting the content script once, each time you open the popup.

wbamberg:
@Rob--W , would you mind taking a look at this version?

beastify/popup/choose_beast.js
```diff
@@ -1,54 +1,81 @@
...
+                  }`;
 
-/*
-Listen for clicks in the popup.
+/**
+ * When the popup loads, inject a content script into the active tab,
+ * and add a clck handler.
+ */
+browser.tabs.executeScript({file: "/content_scripts/beastify.js"})
+.then(() => {
```
Rob--W: 1. This approach won't work if the tab unloads/reloads while the popup has been opened.
2. And if the extension is unable to access the document (e.g. privileged URLs such as `about:preferences`), then the buttons won't do anything upon "click".
3. Lastly, if the page is still loading, then the content script won't be injected (until the document has finished, because `runAt` is not set, so it defaults to `document_idle`) and the button will also appear to not handle clicks.

The second issue should be fixed, e.g. by adding a second callback to `.then` and then updating the UI as appropriate.
To keep the example simple you can just document the other limitations (the situations are quite rare anyway).

wbamberg:
@Rob--W , I've made these updates, please take another look.

beastify/popup/choose_beast.js
```diff
@@ -77,5 +76,22 @@ browser.tabs.executeScript({file: "/content_scripts/beastify.js"})
...
   });
-  
-});
+}
+
+/**
+ * There was an error executing the script.
+ * Display the popup's error message, and hide the normal UI.
+ */
+function reportExecuteScriptError() {
```
Rob--W: Not: Maybe add an error parameter and then

    console.error(`Failed to beastify: ${errror.message}`);

so that beginners can know how to see error messages for debugging?

wbamberg:
Thanks for all your reviews and help on this @Rob--W . I'm really happy to get it merged :).

PR #246, opened 2017-07-13
List delete cookies

ericjung:
Add cookie deletion capabilities.


andymckay:
Unfortunately it looks like the eslint changes hit this, you'll need to rebase this @ericjung.

PR #247, opened 2017-07-14
test travis

andymckay:
do not merge, this is to test that travis ci is set up for pull requests and enforces the code that @yfdyh000 set up

### DIFF
--- a/test-travis/foo.js
+++ b/test-travis/foo.js
@@ -1,0 +1,1 @@
+if (1 == 0) { console.log('bad') }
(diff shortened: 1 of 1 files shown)


andymckay:
@wbamberg I'm not an admin on this repo, I think you will need to set up travis ci so it runs.

wbamberg:
@andymckay , can you check if it's running now?

andymckay:
Hmm, doesn't seem to be.

andymckay:
Spoke to soon, totally worked.

We should try and clean up those warnings at some point.

Thanks @wbamberg and @yfdyh000

wbamberg:
Yes, thanks @yfdyh000 !

PR #251, opened 2017-07-18
Selfify example

rebloor:
Example to accompany an article on the methods of working with files in an extension. This example creates a sidebar on which there is a file picker and drag and drop zone. When an image file is selected in the sidebar it's used to replace the content of the current page.


selfify/content_scripts/selfify.js
```diff
@@ -0,0 +1,41 @@
+/*
+"Selfifies" the web page by:
+* Removing every node in the document.body
+* Inserting the selected image
+* Removing itself as a listener
+*/
+function selfify(request, sender, sendResponse) {
+  debugger;
```
andymckay: Should probably remove this.

selfify/manifest.json
```diff
@@ -0,0 +1,23 @@
...
+  "description": "Adds a sidebar offerin a file picker and drap and drop zone. When an image file is chosen the active tab's body content is replaced with file selected. See https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Examples#selfify",
+  "manifest_version": 2,
+  "name": "Selfify",
+  "version": "1.0",
+  "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/selfify",
+
+  "permissions": [
+    "activeTab",
+    "tabs",
+    "<all_urls>"
```
andymckay: Do you need both activeTab and <all_urls>?

selfify/sidebar/choose_file.js
```diff
@@ -0,0 +1,75 @@
+
```
andymckay: Extra new line.

selfify/sidebar/choose_file.js
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+  // Get the active tab for the sidebar
+  await browser.tabs.query({windowId: myWindowId, active: true})
+    .then((tabs) => {
+      tabId = tabs[0].id;
+    });
+
+  // Inject the content script into the active tab
+  console.log("TAB ID", tabId);
```
andymckay: Maybe something like "Selfify injecting into tab id:", just so its clearer where its coming from.

selfify/sidebar/choose_file.js
```diff
@@ -0,0 +1,75 @@
...
+  // Get the Window ID.
+  await browser.windows.getCurrent({populate: true}).then((windowInfo) => {
+    myWindowId = windowInfo.id;
+  });
+
+  // Get the active tab for the sidebar
+  await browser.tabs.query({windowId: myWindowId, active: true})
+    .then((tabs) => {
+      tabId = tabs[0].id;
+    });
```
andymckay: I think you can just pass windows.WINDOW_ID_CURRENT to tabs.query and remove the windows lookup? Or is that different.

selfify/sidebar/sidebar.css
```diff
@@ -0,0 +1,6 @@
+
```
andymckay: spare new line

selfify/sidebar/sidebar.css
```diff
@@ -0,0 +1,6 @@
...
+.drop_zone {
```
andymckay: I think this should be # because its an id, not a class, in the HTML. The drop zone right now is really small.

selfify/sidebar/choose_file.js
wbamberg: Yeah, that seems to work. You can also do `browser.tabs.query({currentWindow: true, active: true})`.

wbamberg: With `await`, you don't need `then`. You can do:

`let tabs = await browser.tabs.query({windowId: browser.windows.WINDOW_ID_CURRENT, active: true});`

selfify/sidebar/sidebar.css
```diff
@@ -0,0 +1,6 @@
...
+  border: 5px solid blue;
+  width:  400px;
+  height: 200px;
```
wbamberg: It would also be good to have some margin for the input element, it's crammed right underneath the title now.

selfify/sidebar/sidebar.html
```diff
@@ -0,0 +1,26 @@
...
+    
+  </head>
+
+<body>
+  <input type="file" accept="image/*" id="input">
+  
+  <p></p>
+  
+  <div id="drop_zone">
+    <strong>Drag an image file into this Drop Zone ...</strong>
```
wbamberg: Again it would be good to give this `<strong>` a bit of breathing room, for instance:

```
#drop_zone_label {
  margin: 1em;
  display: block;
}
```

selfify/viewer.html
```diff
@@ -0,0 +1,20 @@
...
+    <style>
+     html, body {
+       margin: 0;
+       padding: 0;
+     }
+
+     img {
+       width: 100%;
+     }
+    </style>
```
wbamberg: Can we have this in an external stylesheet?

selfify/content_scripts/selfify.js
```diff
@@ -0,0 +1,41 @@
...
+
+/*
+Given a URL to an image, create and style an iframe containing an 
+IMG node pointing to that image, then insert the node into the document.
+*/
+function insertImage(imageURL, fileName) {
+  var insertImage = document.createElement("iframe");
+  insertImage.setAttribute("src", browser.extension.getURL(`/viewer.html?blobURL=${imageURL}`));
+  insertImage.setAttribute("style", "width: 100vw");
+  insertImage.setAttribute("style", "height: 100vh");
```
wbamberg: This overwrites line 30, so you end up only setting height, which I think might be why the image ends up tiny.

I think you need something more like:

`insertImage.setAttribute("style", "width: 100vw; height: 100vh;");`

selfify/content_scripts/selfify.js
```diff
@@ -0,0 +1,41 @@
...
+  document.body.appendChild(insertImage);
+  var info = document.createElement("span");
+  info.innerHTML = fileName;
```
wbamberg: Please don't use `innerHTML`: https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#Security_considerations.

Although as I said, I'd be happy to not do the thing with filename, and just set the image.

selfify/content_scripts/selfify.js
rebloor: Hi Andy, you meant removing debugger; right?

rebloor:
Hi Will, with the exception of possibly renaming the example, I think all the other changes are done, sorry there were so many. If everything is OK I'll rename the example.

selfify/content_scripts/selfify.js
andymckay: Yes.

selfify/content_scripts/selfify.js
```diff
@@ -0,0 +1,49 @@
...
+    while (document.body.firstChild) {
+      document.body.firstChild.remove();
+    }
+  }
+
+  /*
+  Given a URL to an image, create and style an iframe containing an 
+  IMG node pointing to that image, then insert the node into the document.
+  */
+  function insertImage(imageURL, fileName) {
```
wbamberg: `fileName` is now unused and should be removed.

selfify/content_scripts/selfify.js
```diff
@@ -0,0 +1,49 @@
...
+    var insertImage = document.createElement("iframe");
+    insertImage.setAttribute("src", browser.extension.getURL(`/viewer.html?blobURL=${imageURL}`));
+    insertImage.setAttribute("style", "width: 100vw; height: 100vh;");
+    document.body.appendChild(insertImage);
+    var info = document.createElement("span");
```
wbamberg: This `<span>` is also vestigial now I think.

selfify/manifest.json
wbamberg: @rebloor , can you confirm Andy's question here?

selfify/sidebar/choose_file.js
```diff
@@ -0,0 +1,63 @@
...
+  e.stopPropagation();
+  e.preventDefault();
+  displayFile(e.dataTransfer.files);
+}
+
+/* 
+Insert the content script and send the image file ObjectURL to the content script using a 
+message.
+*/ 
+async function displayFile(fileList) {
```
wbamberg: This doesn't need to be `async` now it doesn't use `await`

selfify/sidebar/choose_file.js
```diff
@@ -0,0 +1,63 @@
+/*
+Listens for a file being selected, creates a ObjectURL for the chosen file, injects a
+content script into the active tab then passes the image URL through a message to the
+active tab ID.
+*/
+
+// Listen for a file being selected through the file picker
+let inputElement = document.getElementById("input");
```
wbamberg: Sorry I should have been clearer. I didn't mean, consider making *this one `var`* a `let` or `const`.

I meant, in general, in this source file, and the others, *consider* whether using `const` or `let` would be better. I think it's OK to use `var` everywhere, so if that's your preference I think it's fine. But `const` or `let` are probably better practice in many situations (personally I would prefer to use `const` in this particular case, where I'm storing a reference to a node). 

However, I don't think changing *just this one* `var` to `let` is an improvement: it's better to have a style and stick to it consistently.

wbamberg:
Thanks for making these updates! I had a few more comments.

> sorry there were so many

No worries, sorry I'm picky.

selfify/content_scripts/selfify.js
rebloor: And from:
_  function selfify(request, sender, sendResponse) {
    removeEverything();
    insertImage(request.imageURL, **request.fileName**);
  }_
too

selfify/sidebar/choose_file.js
rebloor: Sorry, seems I was behind on ES6+ best practices :-( On review, const seems the best for all.

rebloor:
If we are all good now - I'll rename, hopefully the final change

selfify/content_scripts/selfify.js
```diff
@@ -0,0 +1,48 @@
...
+  /*
+  Given a URL to an image, create and style an iframe containing an 
+  IMG node pointing to that image, then insert the node into the document.
+  */
+  function insertImage(imageURL) {
+    const insertImage = document.createElement("iframe");
+    insertImage.setAttribute("src", browser.extension.getURL(`/viewer.html?blobURL=${imageURL}`));
+    insertImage.setAttribute("style", "width: 100vw; height: 100vh;");
+    document.body.appendChild(insertImage);
+    document.body.appendChild(info);
```
wbamberg: This line still needs to be removed.

selfify/manifest.json
rebloor: Don't need activeTab - removed

selfify/content_scripts/selfify.js
rebloor: Removed

rebloor:
Last nits (I hope) and a rename

rebloor:
Rats, you are right, and so much for my folder wide text find :-( (Although I should have known, workman, tools and all that)

wbamberg:
Thanks @rebloor!

rebloor:
Thanks @wbamberg

PR #260, opened 2017-08-10
Update `proxy-blocker` extension to be compatible with firefox 56+

phil-r:
According to [this note](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/proxy/register#compatNote_1) `browser.proxy.registerProxyScript` was changed to `browser.proxy.register`.

I've tested and in firefox 56 `browser.proxy.registerProxyScript` doesn't work anymore

### DIFF
--- a/examples.json
+++ b/examples.json
@@ -315,7 +315,7 @@
         "javascript_apis": [
             "extension.getURL",
             "proxy.onProxyError",
-            "proxy.registerProxyScript",
+            "proxy.register",
             "runtime.onMessage",
             "runtime.sendMessage",
             "storage.local",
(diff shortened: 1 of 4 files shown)


wbamberg:
Thanks! This is a good fix.

There's one other change that should be made here, actually. Due to an old bug, `DIRECT` has to take a dummy argument, here: https://github.com/mdn/webextensions-examples/blob/master/proxy-blocker/proxy/proxy-script.js#L4. That old bug is now fixed, so we see an error in the browser console whenever we request a page that isn't blocked:

`Proxy error: FindProxyForURL: Invalid argument for proxy type: "DIRECT"`

The fix is just to remove the dummy "1234" argument. While you are in there, would you mind making that fix, too?

phil-r:
@wbamberg no problem ;)
I removed a dummy argument

PR #264, opened 2017-08-14
change shortcut to Ctrl+Shift+U for commands example

zlatin:
Fix for keybord shortcut description to Ctrl+Shift+U on [MDN Example extensions page](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Examples).

### DIFF
--- a/examples.json
+++ b/examples.json
@@ -82,7 +82,7 @@
             "commands.onCommand"
         ],
         "name": "commands",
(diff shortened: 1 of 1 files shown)


mmuman:
ctrl-shift-u is used on GTK+ based desktops (GNOME) at least on GNU/Linux to enter Unicode symbols…

PR #269, opened 2017-08-20
Add metadata-checker example

mmuman:
Shows OpenGraph/Twitter/Facebook (minimalistic) previews and metadata.

Demonstrates passing a return value from the content script, to extract data from the page DOM.


wbamberg:
Thanks, this is a neat example. I'm no expert in social meta tags, so some of these comments might be wrong.

This doesn't work with the Twitter metadata, at least for the sites I've tried:
https://developer.mozilla.org/
https://css-tricks.com/

I think this is because those sites use `<meta name="twitter:card"` where the content script is looking only for `<meta property=`. The sites seem to be doing the right thing according to the [documentation I've found](https://css-tricks.com/essential-meta-tags-social-media/), and according to [Twitter's own validator](https://cards-dev.twitter.com/validator).

Even with that fixed though, it doesn't show the image, I think because the popup code is looking for [`twitter:image:src` rather than just `twitter:image`](https://github.com/mdn/webextensions-examples/pull/269/files#diff-bcaadab486d71c38ba1068be3794eff3R67).

It seems [Facebook uses OpenGraph tags](https://developers.facebook.com/docs/sharing/webmasters), so I would just omit all the Facebook-specific code and only show OpenGraph and Twitter. 

Personally, I'd also not bother [showing all the raw data](https://github.com/mmuman/webextensions-examples/blob/667e9f0cb93c85f96cc61cd45bb655a7c6f7eacd/metadata-checker/popup/metadata.js#L5-L22) - I'd just show the rendered version. Along with that, I'd also not bother converting the JSON you get from the content script into that [`data`](https://github.com/mmuman/webextensions-examples/blob/667e9f0cb93c85f96cc61cd45bb655a7c6f7eacd/metadata-checker/popup/metadata.js#L3) object, I'd just traverse the JSON immediately to create the popup's DOM.

I'd also omit a section entirely if the page doesn't have any metadata of that type. It looks weird to have just an H1 heading with no content. (speaking of sections, maybe you could use a [`<section>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section) for each type).

mmuman:
Thanks for the comments.

I didn't check the secs for name vs property, I guess I'll just catch both.

It seems they deprecated the twitter:image:src & other things in favor of just specifying the url directly in twitter:image, as it's [the only one documented now](https://dev.twitter.com/cards/markup); but some sites still use that. It should fall back to it but support both.

As for Facebook, indeed they use og: directly. I kept the separation to list the fb properties like app_id.

I know the meta content is available in the page info context menu, but as a developer I'd like to have them handy along with the preview, as the preview itself might be wrong (it doesn't check the twitter:card, ...).
I think I'll group them at the end of the page instead.

Btw, I'm thinking about using the SVG directly for the icon, but [I don't see any mean to specify both](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/icons).

mmuman:
Any other thing you wanted to change?

wbamberg:
Sorry @mmuman , I have taken another look, but didn't yet get time to write up my comments. But it is on my list I promise.

mmuman:
No pressure, I have so many other things to attend actually anyway, just eager to check things on my TODO list :)

metadata-checker/README.md
```diff
@@ -0,0 +1,9 @@
+# metadata-checker
+
+## What it does
+
+This example adds a popup that lets check the opengraph, twitter and facebook metadata from the current web page.
```
wbamberg: Please use correct syntax and capitalization for things like Open Graph, Twitter, Facebook (and it would be helpful to link to the docs for their metadata too).

metadata-checker/content_scripts/get_metadata.js
```diff
@@ -0,0 +1,16 @@
+var metaData = {};
+
+/* Parse the meta tags to find OpenGraph and Twitter data. */
+/* TODO: OG allows multiple og:image, while JS object properties must be unique... */
+for (meta of document.getElementsByTagName("meta")) {
```
wbamberg: -> `for (let meta of`

metadata-checker/manifest.json
```diff
@@ -0,0 +1,20 @@
+{
+
+  "manifest_version": 2,
+  "name": "Metadata checker",
+  "description": "Displays a popup that summarizes opengraph & twitter metadata from web pages.",
```
wbamberg: Again, please use proper syntax and capitalization in the description.

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
+function onMetaData(metaData) {
+  metaData = metaData[0];
+
+  var shown = false;
+
+  for (prop in metaData) {
```
wbamberg: -> `for (let prop`

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+    value = metaData[prop];
+    var section = null;
```
wbamberg: This variable is not used.

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+    if (!prop.match(/^og:/) && !prop.match(/^twitter:/) && !prop.match(/^fb:/))
+      continue;
+    if (!shown) {
```
wbamberg: I think this is always `true`, making `shown` redundant.

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+      document.getElementById("metadata").setAttribute("class", "");
+      document.getElementById("nothing-to-see").setAttribute("class", "hidden");
+    }
+    let ul = document.getElementById("metadata-list");
```
wbamberg: You sometimes use `let` and sometimes `var`, without an obvious reason (not obvious to me anyway :). Try to be consistent throughout the extension (some people seem to still use `var` everywhere, which is fine IMO, but if you use other forms, please use them consistently).

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+    let li = document.createElement("li");
+    let tt = document.createElement("tt");
```
wbamberg: Please don't use `tt`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+    li.appendChild(tt);
+    li.appendChild(document.createTextNode(": "+ value));
+    ul.appendChild(li);
+  }
+
+  // see http://ogp.me/
+  if (metaData["og:url"] || metaData["og:title"] || metaData["og:description"]) {
+    let container = document.getElementById("og-preview");
+    container.setAttribute("class", "");
+    //XXX: handle arrays ?
```
wbamberg: Again, remove TODO-type comments

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+    var image = metaData["og:image"] || metaData["og:image:src"];
+    // TODO: fixup "//..." urls
+    // cf. https://css-tricks.com/essential-meta-tags-social-media/
```
wbamberg: Here again

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+    //if (image && image.match(/^\/\/:/))
+    //  image = (new URL(tabUrl)).protocol + image;
```
wbamberg: No commented-out code please.

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+    if (image) {
+      let div = document.createElement("div");
+      let img = document.createElement("img");
+      img.setAttribute("style", "max-width: 100px; max-height: 100px;");
```
wbamberg: I'm not sure why you are setting this as an inline style, rather than in the stylesheet.

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+      //TODO: check og:image:width & height
```
wbamberg: Another TODO

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  // see https://dev.twitter.com/cards/markup
+  if (metaData["twitter:card"] || metaData["twitter:description"] || metaData["twitter:title"]) {
+    let container = document.getElementById("tw-preview");
+    container.setAttribute("class", "");
+
+    var card = metaData["twitter:card"];
+
+    // "then a summary card may be rendered."
```
wbamberg: Context of this comment is not obvious

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+      div.setAttribute("class", "warning");
+      div.appendChild(document.createTextNode("Missing twitter:card!"));
+      container.appendChild(div);
+    }
+
+    let image = metaData["twitter:image"] || metaData["twitter:image:src"] || metaData["og:image"] || metaData["og:image:src"];
+    // TODO: fixup "//..." urls
+    //if (image && image.match(/^\/\/:/))
+    //  image = new URL(tabUrl).protocol + image;
+    if (image) {
```
wbamberg: More TODO and commented-out code.

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+      let div = document.createElement("div");
+      let img = document.createElement("img");
+      if (card == "summary_large_image")
+        img.setAttribute("style", "max-width: 300px; max-height: 300px;");
```
wbamberg: Please use brackets even for 1-liners.

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+    if (description) {
+      let content = document.createTextNode(description);
+      div.appendChild(content);
+    }
+    container.appendChild(div);
+  }
+}
+
+function showMetaDataForTab(tabs) {
+  //get the first tab object in the array
```
wbamberg: Please have a space between `//` and text.

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
...
+
+  var gettingMetaData = browser.tabs.executeScript(tab.id, {
+    file: "/content_scripts/get_metadata.js"
+  });
+  gettingMetaData.then(onMetaData);
+}
+
+//get active tab to run an callback function.
+//it sends to our callback an array of tab objects
+function getActiveTab() {
```
wbamberg: Also here, and the language could be clearer, for example:

```
// Get the active tab and pass it to showMetaDataForTab.
// The tab is passed as an array containing a single Tab object.
```

metadata-checker/popup/metadata.js
```diff
@@ -0,0 +1,138 @@
+function onMetaData(metaData) {
+  metaData = metaData[0];
```
wbamberg: It's probably worth commenting this line, e.g.

```
Because tabs.executeScript() may inject the script into more than one frame,
it is resolved with an array of objects, one for each frame in which it ran,
representing the result of executing the script in that frame.
```
...or something like that.

metadata-checker/popup/metadata.js
wbamberg: I just realised as well: you can simplify all this, because if you omit the tabId argument to `tabs.executeScript` then the script is injected into the active tab: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/executeScript#Parameters.

mmuman:
Hi, sorry I was quite busy, but at least I'm sleeping better now...
I'll look into all this.

metadata-checker/content_scripts/get_metadata.js
```diff
@@ -0,0 +1,16 @@
...
+  var prop = meta.getAttribute("property");
+  if (!prop)
+    prop = meta.getAttribute("property");
```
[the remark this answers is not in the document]
mmuman: Looks like I was really asleep when I copy-pasted this 😁
Right the boolean operator makes it simpler indeed.

metadata-checker/popup/metadata.js
mmuman: I actually always used var up to now, but somewhere else I saw let and var being used and I somehow got the impression that let was kind of a "const" or something, but I didn't really ever looked into the differences. Now [I get it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let) :D

mmuman:
I'm afraid it'll take some more time, lots of events coming this months.

PR #272, opened 2017-08-25
Update context-menu-demo

wbamberg:
This updates the context-menu-demo example to reflect changes which landed in https://bugzilla.mozilla.org/show_bug.cgi?id=1268020 and https://bugzilla.mozilla.org/show_bug.cgi?id=1333403.

* rename from context-menu-demo to menu-demo
* use "menus" as the namespace and permission string, rather than "contextMenus"
* add a new demo of the "tools_menu" context


wbamberg:
Thanks @andymckay , I've added a note to the README about this.

PR #278, opened 2017-09-01
example to Demonstarate usage of pageAction API

tsl143:
A standalone example to demonstrate the usage of pageAction API.


wbamberg:
This is a neat idea for an example, and I'm sorry to not merge it.

However, we already have [three examples that demonstrate the pageAction API](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/pageAction#Example_extensions) and it's not really clear to me what this one adds. There are a lot of examples in this repo now, and I'd prefer we focus on examples that demonstrate features that aren't already covered.

andymckay:
It might have been to demonstrate something on Android. @tsl143 do the other examples work on Android?

PR #281, opened 2017-09-08
add in simple http-response example

andymckay:
* the new and shiny from bug https://bugzilla.mozilla.org/show_bug.cgi?id=1255894


wbamberg:
This looks good, but...
(1) it won't install in Nightly, because Firefox Nightly is "57.0a1", which is considered earlier than the "57" specified in strict_min_version
(2) with that fixed, Nightly (clean profile) reliably crashes when I visit https://example.com

andymckay:
@wbamberg 

1) ok will fix.
2) eek, can we get a bug on that in bugzilla ASAP for Kris's attention, we don't have much time to fix that.

andymckay:
@wbamberg https://bugzilla.mozilla.org/show_bug.cgi?id=1399005 was fixed, so if you could retry tomorrow maybe, that would be cool.

PR #284, opened 2017-09-13
create a find example

andymckay:
* an example to demonstrate the find API


andymckay:
Ping @wbamberg? Was going to add to the 57 blog post.

.eslintrc.json
```diff
@@ -1,7 +1,7 @@
 {
   "root": true,
   "parserOptions": {
-    "ecmaVersion": 6
+    "ecmaVersion": 2017
```
wbamberg: I'm not sure what this is for, but if it's intentional I'll take your word for it.

find-across-tabs/manifest.json
```diff
@@ -0,0 +1,30 @@
+{
+  "description": "Find across all your tabs",
+  "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/find-across-tabs",
+  "manifest_version": 2,
+  "name": "Find across tabs",
+  "permissions": [
+    "find",
+    "tabs",
+    "<all_urls>"
```
wbamberg: It seems you don't need `<all_urls>`.

.eslintrc.json
andymckay: Otherwise eslint is not happy with async keyword and I think async await keeps this code cleaner.

find-across-tabs/manifest.json
andymckay: Oh yeah, good point, will remove.

andymckay:
Changes as requested.

PR #287, opened 2017-09-29
add in managed example

andymckay:


wbamberg:
@andymckay , I've updated this example to point to the docs. Let me know if you are happy with it and I'll merge it.

PR #292, opened 2017-10-02
Check config on Windows

andymckay:
Not sure if we want this or not, but it checks the chain from registry key through to batch script. This can maybe help people with the hoops that they have to jump through. It outputs something like this:

```
C:\Users\andy\sandboxes\webextensions-examples\native-messaging\app>python check_config_win.py
Checking: HKEY_CURRENT_USER Software\Mozilla\NativeMessagingHosts\ping_pong
Path from registry key is: c:\users\andy\sandboxes\webextensions-examples\native-messaging\app\ping_pong.json
Path from JSON is: C:\Users\andy\sandboxes\webextensions-examples\native-messaging\app\ping_pong_win.bat
Path from batch file is: C:\Users\andy\sandboxes\webextensions-examples\native-messaging\app\ping_pong.py
Looks good! Give it a try from Firefox.
```


native-messaging/app/ping_pong.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ping_pong",
   "description": "Example host for native messaging",
-  "path": "/path/to/native-messaging/app/ping_pong.py",
+  "path": "C:\\Users\\andy\\sandboxes\\webextensions-examples\\native-messaging\\app\\ping_pong_win.bat",
   "type": "stdio",
   "allowed_extensions": [ "ping_pong@example.org" ]
 }
```
wbamberg: I think this change is unintentional?

native-messaging/app/ping_pong_win.bat
```diff
@@ -1,3 +1,3 @@
 @echo off
 
-call python C:\path\to\ping_pong.py
+call python C:\\Users\\andy\\sandboxes\\webextensions-examples\\native-messaging\\app\\ping_pong.py
```
wbamberg: I think this change is unintentional too?

andymckay: Umm yeah, I'll remove those changes.

andymckay:
I'll move it to root directory too.

andymckay:
Sorry for the delay, updated with feedback.

native-messaging/check_config_win.py
```diff
@@ -0,0 +1,64 @@
...
+
+Currently requires Python 3.
+
+If you find more issues with setting this up, let's see if we can add to this
+script.
+'''
+import json
+import os
+import re
+import winreg
```
pmario: hi, I do get: 

```
PS D:\git\webextensions-examples\native-messaging> python.exe .\check_config_win.py
Traceback (most recent call last):
  File ".\check_config_win.py", line 13, in <module>
    import winreg
ImportError: No module named winreg
```

IMO some more info is needed about dependnecies.

andymckay: winreg is a standard Python 3 module, are you using Python 2? If so making this script work in Python 2 could be a seperate bug: https://docs.python.org/3/library/winreg.html

pmario: did install python3, but still an error. ... I don't have access to win machine atm. so I'll post again tomorrow.

native-messaging/check_config_win.py
```diff
@@ -0,0 +1,64 @@
...
+
+key_path = 'Software\\Mozilla\\NativeMessagingHosts\\ping_pong'
+# Assuming current user overrides local machine.
+key_roots = ['HKEY_CURRENT_USER', 'HKEY_LOCAL_MACHINE']
+
+found_key = False
+
+for root in key_roots:
+    key = winreg.OpenKey(getattr(winreg, root), key_path)
```
pmario: Using python 3.x I get this info. 
```
PS D:\git\webextensions-examples\native-messaging> python .\check_config_win.py
Traceback (most recent call last):
  File ".\check_config_win.py", line 22, in <module>
    key = winreg.OpenKey(getattr(winreg, root), key_path)
FileNotFoundError: [WinError 2] Das System kann die angegebene Datei nicht finden
```

which is ok. Since I didn't create a registry key yet. .... BUT IMO the error message is misleading.

What I would expect is something similar to: 

```
Please create a registry key as described in: https://developer.chrome.com/extensions/nativeMessaging#native-messaging-host-location  ... BUT ... use the key_path XXXXXX
```

Or something that works. ... The problem I have atm ... I don't have an idea, what's really needed! and the example docs doesn't help. .. and the link: https://developer.chrome.com/extensions/nativeMessaging#native-messaging-host-location is kind of crap too. since it contains chrome info and not FF info.

wbamberg: Did you see: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Native_manifests#Manifest_location?

pmario: nope. thx

pmario: But still. I think that link should be returned then :)

andymckay: Please open an issue on this repo so then a contributor can maybe fix it.

PR #300, opened 2017-10-13
Added a new example, session-state

wbamberg:
This adds an example of using `sessions.setTabValue` and `sessions.getTabValue` to store and retrieve state that the extension wants to associate with a tab. This means the extension can restore this state if the user closes and then restores the tab.

I'd like feedback on whether this is a good example of how to use the API, as well as whether the code seems to be correct. I'm not very keen on the onCreated->onUpdated stuff I do here, but it seems that onCreated is too early to call insertCSS. Of course there are alternative ways we could illustrate the concept without needing insertCSS, but they bring they own complexities too. But I'd be happy to hear suggestions for simplifying the example.


wbamberg:
@bobsilverberg , since you reviewed the code for these APIs, I wondered if you would mind reviewing this example?

session-state/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+## What it does
+
+This extension adds a context menu item. When the user clicks the context menu item, the extension adds a border to the page.
+
+But the extension also uses `sessions.setTabValue` to store the fact that it has added a border to this page. If the user closes the page, then restores it, the extension will retrieve this fact using `sessions.getTabValue`, and use that to reapply the border.
+
+## What it shows
+
+This example demonstrates how you can use the [sessions](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/sessions) API to store and retrieve arbitrary state that you want to associate with a tab or window. Then if the tab/window is closed and subsequently restored, you can retrieve the state.
```
andymckay: I would maybe add some notes about how to do a restore in firefox, as mentioned on IRC so its clear the difference between that and just opening example.com again :)

session-state/background.js
```diff
@@ -0,0 +1,62 @@
+/**
+ * Add a context menu item to borderify the current tab.
+ */
+browser.menus.create({
+  id: "borderify",
```
andymckay: minor nit, but if someone had this and borderify installed this might be confusing

session-state/background.js
```diff
@@ -0,0 +1,62 @@
...
+/*
+browser.tabs.onUpdated.addListener((tabId) => {
+  console.log(tabId);
+  browser.sessions.getTabValue(tabId, "border-css").then(result => {
+    if (result) {
+      browser.tabs.insertCSS(tabId, {code: result, runAt: "document_end"});
+    }
+  });
+});
+*/
```
andymckay: just remove this?

bobsilverberg:
It looks like a decent example to me, @wbamberg. Thanks for creating it.

Issue #36, opened 2016-02-04
Have coverage for the 50 most popular APIs in the Chrome store


wbamberg:
Here's the list, courtesy of andym. I'll keep this report updated as we add new examples. When all entries are "done" or "not doing", this bug can be closed.

Of course, a single example add-on could (and probably should) demonstrate more than one API. So this does not necessarily mean that we need 50 examples.

**Remaining:**
```
chrome.runtime.onInstalled
chrome.runtime.getManifest
chrome.runtime.reload
chrome.webNavigation.onCommitted
chrome.webNavigation.onBeforeNavigate
```

**Done:**
```
chrome.browserAction.onClicked
chrome.browserAction.setBadgeText
chrome.browserAction.setBadgeBackgroundColor
chrome.browserAction.setTitle
chrome.browserAction.setIcon
chrome.contextMenus.create
chrome.contextMenus.onClicked
chrome.extension.getURL
chrome.extension.getBackgroundPage
chrome.i18n.getMessage
chrome.management.getAll
chrome.notifications.create
chrome.runtime.sendMessage
chrome.runtime.onMessage
chrome.runtime.connect
chrome.runtime.onConnect
chrome.storage.sync
chrome.storage.local
chrome.tabs.query
chrome.tabs.create
chrome.tabs.sendMessage
chrome.tabs.executeScript
chrome.tabs.onUpdated
chrome.tabs.update
chrome.tabs.onRemoved
chrome.tabs.reload
chrome.tabs.onCreated
chrome.tabs.get
chrome.tabs.onActivated
chrome.tabs.remove
chrome.windows.update
chrome.windows.getCurrent
chrome.windows.getAll
chrome.windows.create
chrome.webRequest.onErrorOccurred
```

**Not doing:**
```
chrome.tabs.getSelected (deprecated)
chrome.tabs.onReplaced (Chrome-only)
```

wbamberg:
Removed deprecated `extension` APIs.

wbamberg:
Updated, marked a couple more APIs as deprecated or unsupported.

andymckay:
The following:

```
chrome.runtime.onInstalled
chrome.runtime.reload
```

Need a fully signed and installed add-on that then updates. We could do it on MDN or set it up all here. Is that worth it?

andymckay:
Actually it might be worth having an add-on on this site that's signed and has an update URL to show how all that works.

andymckay:
[...]

PR #304, opened 2017-10-26
add in a runtime example

andymckay:
Coverage for:

```
chrome.runtime.onInstalled
chrome.runtime.getManifest
chrome.runtime.reload
```

From #36


runtime-examples/manifest.json
```diff
@@ -0,0 +1,17 @@
+{
+  "browser_action": {
+    "default_icon": "run.png"
+  },
+  "description": "Some runtimes example",
+  "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/tabs-tabs-tabs",
```
wbamberg: This should be updated.

runtime-examples/manifest.json
```diff
@@ -0,0 +1,17 @@
+{
+  "browser_action": {
+    "default_icon": "run.png"
+  },
+  "description": "Some runtimes example",
```
wbamberg: Maybe something a bit more complete, like "Demonstration of some runtime APIs."

runtime-examples/manifest.json
```diff
@@ -0,0 +1,17 @@
...
+  "homepage_url": "https://github.com/mdn/webextensions-examples/tree/master/tabs-tabs-tabs",
+  "manifest_version": 2,
+  "name": "Runtime examples",
+  "version": "1.0",
+  "permissions": ["notifications"],
+  "background": {
+    "scripts": ["background.js"]
+  },
+  "icons": {
+    "128": "run.png"
```
wbamberg: great icon :)

runtime-examples/background.js
```diff
@@ -0,0 +1,21 @@
...
+    type: 'basic'
+  });
+}
+
+function onClick() {
+  browser.runtime.reload();
+}
+
+browser.browserAction.setBadgeText({
+  text: `${loadTime.getMinutes()}:${loadTime.getSeconds()}`
```
wbamberg: This gets truncated at 4 characters, so it's confusing (It shows something like "47.2"). I don't think you really need this loadTime/badge text stuff, do you? I'd be tempted to omit it.

andymckay: Really? For me it shows all 5 characters. I wonder why.

We don't really, I just wanted to show the effect of reload in some way. I could add it to the notification message instead where it won't be truncated.

runtime-examples/manifest.json
andymckay: Oops, thanks.

runtime-examples/background.js
wbamberg: That is weird.

>  I just wanted to show the effect of reload in some way.

I think seeing the notification again works for that, doesn't it?

andymckay: It shows that the install notification has been called, but not that the background page (and other stuff) has been reloaded as well.

PR #305, opened 2017-10-27
add in navigation types

andymckay:
* adds in ```chrome.webNavigation.onCommitted``` which helps #36, but I have to confess I can't see an example for onBeforeNavigate
* felt that some of the strings were a bit long and so did some "cleaning"


andymckay:
@wbamberg ping

andymckay:
before:

screenshot 2017-12-08 09 12 00

after:

screenshot 2017-12-08 09 12 37

PR #323, opened 2017-12-08
Use for...in instead of for...of

mmarzantowicz:
This example doesn't work as expected - it does not clear the history.
`for...of` on `Array` returns its elements but array indexes are required, which could be done by `for...in` statement.


andymckay:
Thank you!

PR #325, opened 2017-12-12
Added an example for contentScripts.register

wbamberg:
@andymckay , @rpl , could you please take a look at this?

It's a very simple example of `contentScripts.register()`.

One thing I found: it seems I have to pass `runAt` or I get an error:

```
'runAt' member of WebExtensionContentScriptInit 'null' is not a valid value for enumeration ContentScriptRunAt.
```

Apart from that, I'd be happy to hear any feedback on whether this is a good example and whether the code is OK.


wbamberg:
@rpl, @andymckay , could one or both of you take a look at this example please?

rpl:
@wbamberg I'm sorry for the late reply, I didn't notice the mention on this pull request in my inbox, I'm going to look into it (and I'm only going to file a bugzilla issue for the runAt issue, I think that it should default to document_idle like the content scripts registered from the manifest.json, or the runAt property has to be mandatory instead of optional).

rpl:
Issue files as [Bug 1429425](https://bugzilla.mozilla.org/show_bug.cgi?id=1429425) (contentScripts.register runAt option should default to document_idle when missing)

user-script/popup/user-script.js
```diff
@@ -0,0 +1,17 @@
...
+const codeInput =  document.querySelector("#code");
+
+const defaultHosts = "*://*.org/*";
+const defaultCode = "document.body.innerHTML = '<h1>This page has been eaten<h1>'";
+
+hostsInput.value = defaultHosts;
+codeInput.value = defaultCode;
+
+async function registerScript() {
+  let background = await browser.runtime.getBackgroundPage();
```
andymckay: This won't run in private browsing mode, but just adding that to the readme might be enough.

user-script/popup/user-script.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const hostsInput =  document.querySelector("#hosts");
+const codeInput =  document.querySelector("#code");
+
+const defaultHosts = "*://*.org/*";
+const defaultCode = "document.body.innerHTML = '<h1>This page has been eaten<h1>'";
```
andymckay: I found that this didn't work on https://mozila.org/ for me, could be the runAt bug. Something like an `alert('Hello from user-script)` worked just fine though.

wbamberg: That's odd, it works for me on mozilla.org (specifically https://www.mozilla.org/en-US/). You do have to reload the page though.

user-script/popup/user-script.js
wbamberg: Hm, it would probably be better to use message-passing then. Nobody reads the docs.

user-script/popup/user-script.js
wbamberg: Does it work on other pages, like developer.mozilla.org?

wbamberg:
@andymckay , I changed it to use messaging, and confirmed that it works in private windows. Please let me know if it still looks good.

andymckay:
Cool, suprisingly few lines of code to change that to messaging.

user-script/popup/user-script.js
andymckay: Hmm, working now. Blame user error, I did something wrong :)

andymckay:
ship it squirrel

bhargavmungara:
hi, 

const defaultCode = "document.body.innerHTML= ''"; 

this line is not loading the image, is there any alternatives to load the image without having to change other parts of content registry example

PR #330, opened 2017-12-27
Update README

AllenFang:
I just found the `webpack-modules` example is only for background script and popup instead content.


wbamberg:
Thanks for this! You are right. But I think that this sentence should now say:

"This example shows how to use a node module in a background script and popup."

AllenFang:
@wbamberg updated,  thank you 👍

PR #333, opened 2018-01-12
Themes themes

hellosct1:
examples themes API


wbamberg:
I'm sorry not to accept this PR, but: we have quite a few examples in this repo now. Adding more examples comes with a cost, that it makes it harder for people to find a relevant example. So I only want to accept new examples that illustrate something not already covered by the existing examples.

In this case the PR seems to consist just of the existing static theme examples ([themes](https://github.com/mdn/webextensions-examples/tree/master/themes)) but converted to use the JS themes API instead. We do already have a couple of examples of the theme API ([dynamic-theme](https://github.com/mdn/webextensions-examples/tree/master/dynamic-theme), [private-browing-theme](https://github.com/mdn/webextensions-examples/tree/master/private-browsing-theme)). These existing example seem to cover more realistic use cases (if you just want to apply the theme to every window, why not just use a static theme?).

A couple of other things, for future reference:

* if you submit a large PR like this (any PR really, but especially a large one) please include a proper description of what's in it and why you think it's a good addition.

* you'll see the "All checks have failed" error on this PR. That's because some automated tests failed. If this happens to a PR it's a good idea to fix them before asking for a review.

PR #334, opened 2018-01-17
update to correct link on network-messaging readme

amiechen:
The current link only shows where the app manifest should be for Chrome, which is a little confusing when following the steps to set up the example app. The updated link shows all app manifest locations for FireFox.


wbamberg:
Thanks for this. These docs were reorganised and the link wasn't updated. Your fix is correct, but while you are in there, would you mind also updating the same "App manifest location" link in the "Mac OS/Linux setup" section?

amiechen:
absolutely! just updated it.

PR #335, opened 2018-01-25
use babel-preset-env instead of babel-preset-es2015

hemant1612:

PR #337, opened 2018-02-07
Update react, webpack and babel packages as well as their configs

irsick:
Updated example packages and build configs to be up to date

React to 16.20.0
Webpack to 3.10.0
Web-ext to 2.3.2

Deprecated **babel-plugin-transform-class-properties** 
and **babel-plugin-transform-object-rest-spread** in favor of **babel-preset-env** to let Babel to automatically [detect necessary plugins](http://babeljs.io/docs/plugins/preset-env/).

This pull requests contains the proposal of https://github.com/mdn/webextensions-examples/pull/335

### DIFF
--- a/react-es6-popup/.babelrc
+++ b/react-es6-popup/.babelrc
@@ -1,6 +1,10 @@
 {
   "presets": [
-    "env",
+    ["env", {
+        "targets": {
+            "firefox": 57
+        }
+    }],
     "stage-2",
     "react"
   ]
--- a/react-es6-popup/.babelrc
+++ b/react-es6-popup/.babelrc
@@ -1,11 +1,7 @@
 {
   "presets": [
-    "es2015",
+    "env",
     "stage-2",
     "react"
-  ],
-  "plugins": [
-    "transform-class-properties",
-    "transform-es2015-modules-commonjs"
   ]
 }
--- a/react-es6-popup/package.json
+++ b/react-es6-popup/package.json
@@ -4,22 +4,20 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "build": "webpack -w -v --display-error-details --progress --colors",
+    "build": "webpack -w --display-error-details --progress --colors",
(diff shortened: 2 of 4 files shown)


wbamberg:
@kumar303 , would you mind taking a look at this one, too?

react-es6-popup/.babelrc
```diff
@@ -1,11 +1,7 @@
 {
   "presets": [
-    "es2015",
+    "env",
```
kumar303: By default, it [looks like](http://babeljs.io/docs/plugins/preset-env/) this will transform everything for general web compatibility, including IE and Safari. I think that would be overkill for a browser extension. What if we limited it to Firefox (and maybe Chrome too), like this?

```json
{
  "presets": [
    ["env", {
      "targets": {
        "firefox": 57
      }
    }]
  ]
}
```

react-es6-popup/package.json
```diff
@@ -4,22 +4,20 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "build": "webpack -w -v --display-error-details --progress --colors",
+    "build": "webpack -w --display-error-details --progress --colors",
```
kumar303: Ha. Good catch.

react-es6-popup/webpack.config.js
```diff
@@ -12,25 +12,22 @@ module.exports = {
   output: {
     // This copies each source entry into the extension dist folder named
     // after its entry config key.
-    path: 'extension/dist',
+    path: path.resolve(__dirname) + '/extension/dist',
```
kumar303: I know the old code was broken on Windows but we should try to make this work on Windows. I think this will do it:

```js
path: path.join(path.resolve(__dirname), 'extension', 'dist'),
```

irsick:
Added requested changes

wbamberg:
Thanks for the contribution @irsick , and for the review @kumar303 !

PR #343, opened 2018-03-01
Add an example for menus.onShown and menus.refresh()

wbamberg:
The README should explain what this new example does.


menu-labelled-open/background.js
```diff
@@ -0,0 +1,31 @@
+const openLabelledId = "open-labelled";
+
+let menuItem = browser.menus.create({
```
andymckay: menuItem is never used, maybe remove assignment?

PR #345, opened 2018-03-08
Make menus calls synch

wbamberg:
This makes the calls to menus.update and menus.refresh synchronous, so we don't need to check that the original menu is still open, as per https://bugzilla.mozilla.org/show_bug.cgi?id=1215376#c182.


wbamberg:
@Rob--W , would you mind taking a look at this?

Rob--W:
LGTM. Don't forget to update the inline code on MDN too.

PR #347, opened 2018-03-22
Demonstrate changing keyboard shortcut values

wbamberg:
This extends the "commands" example to demonstrate the `browser.commands.update()` and `browser.commands.reset()` functions added by https://bugzilla.mozilla.org/show_bug.cgi?id=1421811.

It adds an options page to the extension: this contains a textbox and 2 buttons, "update" and "reset". Clicking "update" sets the textbox value as the extension's keyboard shortcut. Clicking "reset" resets it to the original value it has in the manifest.


mstriemer:
Looks good, thanks!

wbamberg:
Thanks!

PR #362, opened 2018-07-11
Add an example using webRequest.getSecurityInfo()

wbamberg:
This is a new example extension that shows how to use [webRequest.getSecurityInfo()](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest/getSecurityInfo). It just collects stats about the root certificates that are used to secure TLS connections, and displays them in a popup.

It's not the most exciting example, but I struggled a bit to think of one simple enough to make a good example but still interesting.

I'd be happy for any feedback on this particular example - or if there's a more interesting possibility, as long as it wouldn't get too complicated.


wbamberg:
@rpl , would you mind taking a look at this?

root-cert-stats/background.js
```diff
@@ -0,0 +1,37 @@
...
+If we haven't seen this subject before, add it. If we have, increment its stats.
+*/
+async function logRootCert(details) {
+  try {
+    let securityInfo = await browser.webRequest.getSecurityInfo(
+      details.requestId,
+      {"certificateChain": true}
+    );
+    if (securityInfo.state == "secure" || securityInfo.state == "weak") {
+      let rootName = securityInfo.certificates[securityInfo.certificates.length - 1].subject;
```
rpl: I think that we may actually want to collect the `issuer` of the last certificate in the chain here (instead of the `subject`).

rpl: How about also keep track of when `securityInfo.isUntrusted  === true`, so that we can make those entries "red colored" once rendered in the popup?

e.g. something like:

```
var rootCertStats = {};
var untrustedCerts = new Set();
...
      let rootName = securityInfo.certificates[securityInfo.certificates.length - 1].issuer;
      ...
      if (securityInfo.isUntrusted) {
        untrustedCerts.add(rootName);
      }
      ...
``` 

and then use that set to highlight the untrusted entries in the popup:

```
    ...
    let entryTR = document.createElement("tr");
    let entryName =  document.createElement("td");
    let entryValue =  document.createElement("td");

    if (backgroundPage.untrustedCerts.has(entry)) {
      entryTR.classList.add("untrusted");
    }
    ...
```

If we add this "feature" to the example, we may also add somewhere in the README.md file a suggestion to navigate a tab on the example pages provided by badssl.com (in particular: https://self-signed.badssl.com/ and https://untrusted-root.badssl.com/), to see the untrusted entries highlighted in the table rendered in the extension popup .

mixedpuppy:
Might be nice to get some input from @april around parsing rawDER.  I suspect that is going to be a popular thing to do.

root-cert-stats/background.js
wbamberg: I am under the impression (and testing seems to confirm this) that the certificate chain includes the trust root itself, which will (always?) be self-signed, so `issuer`===`subject`. In which case it's moot, but I thought `subject` seemed semantically better (subject is the entity associated with the certificate, so for a root cert, it's the trusted organization).

wbamberg: One practical difficulty with this is that (I think) at present we can't get security information if the handshake fails (see https://bugzilla.mozilla.org/show_bug.cgi?id=1322748#c116 and https://bugzilla.mozilla.org/show_bug.cgi?id=1474657).

But even if we could, I'm not sure if this is a good plan. Currently this shows which trust roots are most used, assuming that if the connection worked then the last cert in `certificates[certificates.length-1]` is a trust root.

But if a chain to a trust root couldn't be constructed, then there isn't a trust root to show. I don't actually know what `certificates[certificates.length-1]` would contain in this case. Perhaps the last cert in the list sent from the server? It seems confusing to list that alongside the trust roots.

What I mean is: if a trust root wasn't found, then `certificates[certificates.length-1]` isn't a potential trust root - it's an intermediate or server cert whose issuer is untrusted.

I think, unless I misunderstand things.

april:
I'm using rawDER over in my certificate viewer app:

https://github.com/april/cert-viewer

It's already roughly at feature parity with the Firefox built-in cert viewer, maybe a bit past it. Still got a ways to go. But DER and extensions are pretty complex, so I don't know how well it will do as a demo.

root-cert-stats/background.js
rpl: The main reason why I though that we should collect the `issuer` here is that the table in the popup is titled "Certification Authorities", and (at least if I recall the right definition) CA is supposed to be the entity that "issues the certificates".

rpl: yeah, the user has to add an exception and reload the page to allow the extension to see the security information in case of failed handshakes (and not all the failed handshakes can be bypassed with an exception actually).

I think that you are right, looking at the code behind the `isUntrusted` property, securityInfo.isUntrusted doesn't seem to be enough to be sure if the issuse was because of an unknown/invalid CA or because of one of the other issues that are going to set it to true (e.g. a missing issuer, inadequate key size etc.)

I'm thinking that it may be more reasonable to do not collect the root certificate from a securityInfo that has `isUntrusted === true`.

wbamberg: > CA is supposed to be the entity that "issues the certificates"

Yes, it does issue certificates, but `subject` is the name of the entity identified by the certificate. So if the last cert in the chain is the trusted root (which I think it is) then the name of the authority owning it is found in the subject, not the issuer.

(In practice, as I say, it's generally moot, because usually root certs are self-issued, so subject===issuer. But if they are not, then we want to use the subject, not the issuer.)

I think. But if I'm wrong about any of this I'd be happy to be corrected.

wbamberg: Yes, that makes sense. I had assumed that if `isUntrusted === true` then `state` would be "broken", but if you visit https://untrusted-root.badssl.com/ and add the security exception, then `state` is apparently marked as "secure".

(Which seems a bit weird, now I think about it. Would it be better to have a new value for "state" like "security-exception"?)

Anyway, I've added a check so we won't collect the root if `isUntrusted === true`.

wbamberg:
The cert-viewer looks great, but yes, I don't think I want to include DER decoding in this example. It would be better for Mozilla to include a separate DER decoder that people could use in their extension.

PR #369, opened 2018-08-20
Add example for menus.getTargetElement API

Rob--W:


menu-remove-element/README.md
```diff
@@ -0,0 +1,27 @@
+# menu-labelled-open
```
wbamberg: Please update the title.

Rob--W: Done. I have also expanded the last line.

PR #377, opened 2018-09-25
move beastify.js to beastify and modify choose_beast.js

tejasri19:


wbamberg:
Thanks for the contribution! But I'm not sure why this change is an improvement, can you explain?

Issue #398, opened 2019-01-04
Remove example extensions that depend on npm


wbamberg:
Several of the examples here pull in dependencies for stuff like react, babel, webpack, eslint. They are useful to show how WebExtensions can integrate these technologies, but I think they are outside the core scope of this repo, which is to show how to use the WebExtension APIs.

They are still useful and if they were zero-cost, then I'd be happy to keep them, but they are a significantly greater maintenance burden than extensions that just use the WebExtension APIs. So I'd propose we just remove these examples.

This applies to:

* https://github.com/mdn/webextensions-examples/tree/master/eslint-example
* https://github.com/mdn/webextensions-examples/tree/master/mocha-client-tests
* https://github.com/mdn/webextensions-examples/tree/master/react-es6-popup
* https://github.com/mdn/webextensions-examples/tree/master/store-collected-images
* https://github.com/mdn/webextensions-examples/tree/master/webpack-modules

chrisdavidmills:
I agree with this.

jwhitlock:
The example in ``react-es6-popup`` is the cause of 5 security alerts for the repository, two of them high severity (thanks @openjck). I'm a fan of removing that one first.

irenesmith:
I agree that the web extensions examples should focus on the WebExtension APIs.

motin:
Removing to decrease maintenance seems like a great idea, but maybe they can still live on in one or many separate repos (which in turn can be archived and read-only in case they have no maintainers), or at least linked to from somewhere? Some are still useful despite being a bit outdated.

kumar303:
These examples are really important because `npm` integration greatly enhances the development experience and most developers are already using it for complex extensions, anyway, especially in the Chrome ecosystem. Integrating with `npm` greatly simplifies dependency management, such as keeping [`webextension-polyfill`](https://github.com/mozilla/webextension-polyfill) up to date. 

As an anecdote, I noticed an extension [in the wild](https://github.com/mozilla/secure-proxy) which suggested an `npm install` step *and* a manual step of installing `web-ext`. This is unnecessary because one could just alias `npm start` to the `web-ext` command as the [`react-es6-popup`](https://github.com/mdn/webextensions-examples/tree/f369ddc6421591c78d978bd9f4d7d23c5da77e2d/react-es6-popup) showed.

[...]

PR #400, opened 2019-01-07
Remove react example

wbamberg:
See https://github.com/mdn/webextensions-examples/issues/398 and in particular https://github.com/mdn/webextensions-examples/issues/398#issuecomment-452095705


David263:
The rationale sounds good to me. As a user, I would vote to merge.

PR #412, opened 2019-05-21
Theme manifest key updates

rebloor:
Changes to account for the fact that "headerURL":, "accentcolor":, and "textcolor": have been replaced by "theme_frame": , "frame":, and "tab_text": respectively.  In addition, removed the empty.png image files, as these are no longer needed. Also added an example of use of theme_inactive to weta_tiled.


rebloor:
@irenesmith could you take a look at these, thanks

rebloor:
Thanks, @irenesmith !

Issue #406, opened 2019-03-06
Update examples of manifest.json for compliance with Firefox 48+


GregDomjan:
Had some confusion and couldn't find doc on Applications manifest entry any longer.

Post issue #47 mozilla specific entry `applications` got renamed to more generic `browser_specific_settings` including in part issue [#addons/61](https://github.com/mozilla/addons/issues/61)

Would be nice to update manifest.json with new key, especially in examples that require a min version after 48

https://wiki.mozilla.org/WebExtensions/Applications
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings

rebloor:
Confirmed that no manifest.json includes the key.

PR #416, opened 2019-06-12
Update manifest key applications to browser_specific_settings

e2e8:
It seems that the manifest key "applications" was changed to "browser_specific_settings"
#406 related issue


e2e8:
#406

PR #417, opened 2019-06-29
Improve accessibility

danielnixon:
https://www.w3.org/TR/WCAG20-TECHS/H44.html


rebloor (approved):
Confirmed clicking the label now activates the control.

rebloor:
Thanks @danielnixon

PR #418, opened 2019-06-30
Improve accessibility

danielnixon:


rebloor (approved):
Confirmed that clicking the label activates the control

PR #423, opened 2019-09-18
Change theme on the hour.

brianscottwilson:
Revised the code so the Theme is checked for a change at the start of every hour.

Simplified some of the logic in determining which theme to display.


brianscottwilson:
I took a look at the JavaScript code and it looks good, but I thought it needed some comments, and the logic could be simplified in some places.  I also saw that the logic for when to switch to the day theme wasn't entirely correct on the time of the switch, and I fixed that.  Lastly, instead of checking the theme every five minutes, I set it so the check for a theme change would happen at the start of every hour.

brianscottwilson:
I've tested this on my own system and haven't seen any errors.  The theme change works as expected.

PR #425, opened 2019-09-20
Adds an example for devtools inspector sidebar

nchevobbe:
This adds a webextension example to add an inspector sidebar showing the data of the current selected node in the markup view

### DIFF
--- a/devtools-inspector-sidebar/README.md
+++ b/devtools-inspector-sidebar/README.md
@@ -1,0 +1,12 @@
+# devtools-panels
+
+**Adds a new sidebar to the developer tools inspector.**
+
+## What it does
+
+This extension adds a new sidebar to the inspector panel.
+It displays the properties of the current selected node in the markup view, using
+`sidebar.setExpression($0)` each time a new node is selected (listener added via
+`browser.devtools.panels.elements.onSelectionChanged`).
+
+To learn more about the devtools APIs, see [Extending the developer tools](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Extending_the_developer_tools).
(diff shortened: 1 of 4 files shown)


andreicristianpetcu:
This is awesome! I had no clue this API existed! Thank you, @nchevobbe

wingman-jr-addon:
@nchevobbe This PR looks pretty neat! @caitmuenster Do you know has anyone had a chance to evaluate this PR yet?

rpl (approved):
Looks great.

I'm not sure if the example.json file is still being used to generate the list of examples at:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Examples

If it is it may be nice if we could also add this example in that example.json file, as the other devtools example here:
- https://github.com/mdn/webextensions-examples/blob/60ce50b10ee66f6d706b0715909e756e4bdba63d/examples.json#L142-L153

@chrisdavidmills do you know if we do still regenerate that MDN page from the example.json file in this repo? (and/or how to make sure that the table in the example page is regenerated).

Anyway, we can take care of that in a follow up, and so in the meantime I'm approving this PR as is.

caitmuenster:
@rpl, I just chatted with @chrisdavidmills and it looks like we'll need to manually update the [examples page](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions) on MDN. Let's go ahead and merge this.

rpl:
> @rpl, I just chatted with @chrisdavidmills and it looks like we'll need to manually update the [examples page](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions) on MDN. Let's go ahead and merge this.

sure, that sounds absolutely ok to me (also, thanks for double-checking that!)

nchevobbe:
Thanks for meeting this folks!

Le jeu. 12 nov. 2020 à 19:45, Luca Greco <notifications@github.com> a
écrit :

PR #426, opened 2019-10-07
User scripts register

irenesmith:
Add a new example for the userScripts API, per [Sprints 2136](https://github.com/mdn/sprints/issues/2136)

New example is in: user-script-register-v2


rpl:
@wbamberg I just noticed this PR and I wanted to chime in and help a bit by providing some additional technical details about the API that may be useful (and feel free to ping me for further details or additional feedback)

> The main differences between the two are:
> 
>     * the new example includes settings for all the matching options (globs, frames, etc). IMO this is not very helpful as it adds a lot of code that's not key to an understanding of the API. But it's OK.
> 
>     * the new example adds a file "customUserScriptAPIs.js", which is specified in the manifest's `user_scripts.api_script` key, so it runs when the user script is injected. It provides shims for the storage `get` and `set` APIs. The default script calls those APIs when it first runs, and injects the results into the page content. But none of this is at all explained in the README.

+1 I do agree with @wbamberg, in the README of this example we should mainly mention the key features of this particular API (and maybe briefly mention that the details not mentioned explicitly are shared between both the userScripts and the contentScripts APIs).

> ...
> This seems to be the key feature of user scripts, but that wasn't clear from reading the docs on MDN. The docs there could benefit from an improved overview explaining the purpose of the API.

:+1: Yes, the key features of the userScripts API are basically the following:

- every script registered by the same extension is isolated in its own "global" (on the contrary all the content scripts are executed in the same shared global, one per webpage window)
- the userScript code doesn't run with the privileges of the extension (on the contrary all the "extension content scripts" have "privileges" inherited from the extension "permissions")
- the userScripts code run doesn't have access on any extension APIs on its own, the extension has to inject the APIs that wants to provide to the userScripts (from the "API script")

> There are also some mistakes in the docs:
> 
> "A userScript must be declared in the extension's manifest file using the "user_scripts" manifest key." I don't think this is true. 
> I think the script that listens for `userScripts.onBeforeScript` must be declared in `user_scripts.api_script`. That script is sometimes called an "APIScript" in these docs, although this term isn't defined here. But user scripts are not declared in the manifest as far as I can tell.
> 

yeah, I confirm that too: the userScripts are not going to be usually packaged in the extension itself, and so they can't be declared in the manifest. The userScripts are going to be registered programmatically by the extension using "userScripts.register" from the background page.

> "Once loaded and registered, the user script is executed automatically on any webpage matched by userScript registered by the same extension, before any matched userScript is executed." I don't understand this. The user script is executed before any matched userScript is executed - this doesn't seem to make sense.
> 
> "Once registered, a user script has access to the beforeScript event" - no, the "APIScript" has access to that event. And the "APIScript" is not registered using `register`, it's declared in the manifest AFAICT.
> 
> We also have a link to https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/APIScript - this isn't a JavaScript API so should not get a page here. Similarly, I think camelcasing "APIScript" as if it's a JS API is confusing, since it isn't. Just "API script" would be better.
> 
> In general there seems to be some confusion between user scripts and "API scripts" in this documentation. I think it would be worth being very clear on the difference.

I agree, the MDN API doc pages related to the userScripts API needs a number of fixes from a technical perspective, and we need to make them much more clear about the differences between the userScripts and the "API script".

Some details that seems to be currently missing from the API docs on MDN may still be in these notes I collected some times ago to help Irene to bootstrap the docs: https://docs.google.com/document/d/1xB_I8D77mFcBt-ZHF_yPvZktnF4-A5Zn-Rq5D85azt0/edit#heading=h.fgvqclsqzpax

e.g. some details about the API script are in the following section: https://docs.google.com/document/d/1xB_I8D77mFcBt-ZHF_yPvZktnF4-A5Zn-Rq5D85azt0/edit#heading=h.obnp1s72m9mx

> ...
> In https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/register, the documentation for `scriptMetadata` says just "a JSON object which contains some metadata properties associated with the registered userScripts". Which makes me wonder what the defined properties are. But actually AFAIK you can put anything here, this is just passed through into the API script's `onBeforeScript` handler. So it would be helpful to say that.
> 

The scriptMetadata is an "opaque" object from the userScripts API perspective, the extension can use it to associate some metadata to that particular registered userScript (e.g. a "userScripts manager extension" would use it to associate to the script a JSON representation of the [userScript Metadata block](https://wiki.greasespot.net/Metadata_Block)), then the same metadata will be available on the "API script" side as part of the `script` parameter received from the onBeforeScript listener (and, as an example, a "userScripts manager extension" may use `script.metadata.grants` to decide if a custom API should be granted or not to that particular userScript)   

> In https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/onBeforeScript, `onBeforeScript` is not an event handler, it's a WebExtensions Event object. It's also not clear in what contexts this event is available (i.e. is it only in API scripts?).

+1 from me too:
- it is not a DOM event handler
- we should make it clear (and visible) that this listener is available to the "API script", not to the extension pages (on the contrary the userScripts.register method is only available in the extension pages and not available to the "API script" or the content scripts)
- we should document the parameters expected to be received from the listener (at least as detailed as the [blogpost](https://blog.mozilla.org/addons/2019/03/26/extensions-in-firefox-67/#userscripts) ;-)), some details about the onBeforeScript listener parameters are in the following section: https://docs.google.com/document/d/1xB_I8D77mFcBt-ZHF_yPvZktnF4-A5Zn-Rq5D85azt0/edit#heading=h.1qna29nj8img
- this API doc page for `onBeforeScript` doesn't seem to be linked to any of the other userScripts API doc pages and neither from the sidebar

irenesmith:
Thank you for commenting! I can really use the help. I will be sure to ask
follow up questions.

wbamberg:
Wow, thanks so much for all the extra info @rpl !

rebloor:
@rpl I'm about to pick this up and hopefully have the issues raised addressed in the next few days. However, I've tried to do some research to find out what user scripts are for—the page notes that they are "a special kind of extension content script" but doesn't explain what this means. On digging around the Internet I found some stuff about Greasemonkey user scripts but I'm left unclear whether that's the same or different technology. Are you aware of, or could you provide, a high-level overview that explains what a developer would use user scripts for? Thanks!

rpl:
> @rpl I'm about to pick this up and hopefully have the issues raised addressed in the next few days. However, I've tried to do some research to find out what user scripts are for—the page notes that they are "a special kind of extension content script" but doesn't explain what this means. On digging around the Internet I found some stuff about Greasemonkey user scripts but I'm left unclear whether that's the same or different technology. Are you aware of, or could you provide, a high-level overview that explains what a developer would use user scripts for? Thanks!

@rebloor Have you also looked to the following wiki page? https://wiki.mozilla.org/WebExtensions/UserScripts

The API drafted in that wiki page may not be in sync with the userScripts API as finalized and landed after the review rounds, but the initial part related to the motivations (and also the "Overall" section) should still be relevant.

In short, the userScript API is meant to be a building block used to implement "userScripts manager"-like features.

The extension that uses the userScripts API (like [FireMonkey](https://addons.mozilla.org/en-US/firefox/addon/firemonkey/), which uses the userScripts API to build a kind of "mini-Greasemonkey") is expected to:

- use this API for running the third party userScripts that the user may ask the extension to register, multiple userScripts registered by the same extension will run in sandboxes isolated from each other (which is not the case for regular content scripts) 
- the extension may (optionally) define an APIScript (which has also the same privileged and Extension APIs available to any content script) if the userScripts expect to have some particular API methods available to them (e.g. an extension like FireMonkey does likely use it to provide the same kind of GM.* methods that Greasemonkey provides, see https://wiki.greasespot.net/Greasemonkey_Manual%3aAPI)

@rebloor Let me know if these details are useful to you to get high-level overview of the motivations and purpose of this API (or if there is still something unclear)

rebloor:
Hi @rpl first thanks for the additional information it was very useful and sorry for the delay in responding, the challenge of other priorities. I'm starting to work through the necessary changes and have updated [userScripts](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts) and [userScripts.register()](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/register). 

On [userScripts.register()](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/register) parameters were documented with the comment "Same as (parameter) in the user_scripts key." However, the user_scripts key doesn't contain documentation for any of these parameters, whereas the documentation of the content_scripts key does. So, a couple of questions:
 
1) have these parameters been missed from the user_scripts key? I'm assuming not, they don't seem relevant to the APIScript.
2) If the parameters haven't been missed from user_scripts, is it acceptable to link to the content_scripts key page for their documentation? There is obviously an advantage to linking in that it keeps one source of truth, but a disadvantage in that the reader has to flip between pages to get the information they need. Any thoughts on the preference?  The [
contentScripts.register()](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/contentScripts/register) uses the link option. (FYI @chrisdavidmills)

The `js` parameter is based on `extensionTypes.ExtensionFileOrCode` and therefore includes the `file` option, however, is this relevant to user scripts?

For `scriptMetadata` is there somewhere with more information on what is should contain, or is it any arbitrary Json user wants to include? Given the comment in the schema about this being "opaque", I assume the letter, any arbitrary Json.

Sorry, there may be more questions as I continue editing.

rebloor:
Hi @rpl  Another question about the `js` parameter, it is currently documented as optional. Is that an error from copying the content over from `contentScript`, would there be any point in specifying `userScriptOptions` with just `matches`?

rebloor:
@rpl  can you clarify this for me:

"The APIScript is executed automatically on any page matched by the userScript.register API called from the same extension. It is executed before the user script is executed.

The userScript API also provides a new event, browser.userScripts.onBeforeScript, which the APIScript can listen for.  It is called right before a matched user script is executed, allowing the APIScript to export custom API methods to the user script."

If the APIScript executes automatically on a page before the user script why can't the custom API methods be exported at that point? Or to put it another way, does the script automatically execute, isn't the purpose of onBeforeScript to provide a place to execute the APIScript?

With the exception of figuring out how to eliminate or move the APIScripts page (the answer to the questions above will help) I believe I've now finished all the necessary edits to the API and manifest pages. I will double check tomorrow, but should you have the chance to take a look and provide feedback that would be appreciated.

rpl:
> Hi @rpl Another question about the `js` parameter, it is currently documented as optional. Is that an error from copying the content over from `contentScript`, would there be any point in specifying `userScriptOptions` with just `matches`?

@rebloor yeah, you are right, that property shouldn't be optional for the userScripts.register (it is optional in contentScripts.register because an extension may register a css only content script, but it doesn't make sense for the userScripts which only support the `js` property).

I just checked and we did marked it as optional in the API schema file (https://searchfox.org/mozilla-central/rev/b243debf6235b050b42fd2eb615fdc729636ca6b/toolkit/components/extensions/schemas/user_scripts.json#38) by mistake, and so technically it is optional at the moment.

I filed [Bug 1605403 - UserScriptOptions js property shouldn't be optional](https://bugzilla.mozilla.org/show_bug.cgi?id=1605403) to fix the schema on the Firefox side, in the meantime I think we should suggest it as optional in the API docs (given that a userScript without any script isn't that useful as you pointed out).

rpl:
> If the APIScript executes automatically on a page before the user script why can't the custom API methods be exported at that point? Or to put it another way, does the script automatically execute, isn't the purpose of onBeforeScript to provide a place to execute the APIScript?

the APIScript is executed when the userScript sandbox have not been created yet, and so there isn't any place to inject the custom API into, until the userScripts sandbox objects are actually created.

The purpose of `userScripts.onBeforeScript` is to provide the userScripts sandbox object (plus the opaque scriptMetadata and some additional helper API methods) to the APIScript, and the APIScript can decide in that listener which custom APIs to inject into the userScript (if any, some userScript may not need any).

It is completely possible that the extension may not be even injecting the same custom APIs in all the userScripts sandbox:
the `opaque` scriptMetadata is meant to allow the extension to associate to the registered userScripts some metadata, which is meant to allow the APIScript to decide which APIs should be injected into a specific userScript sandbox (e.g. by parsing the [Greasemonkey metablock](https://wiki.greasespot.net/Metadata_Block), @grant is how a userScript asks the userScript manager extension to make available to the script some specific API, before registering the userScript and turning it into the `scriptMetadata` object) 

> With the exception of figuring out how to eliminate or move the APIScripts page (the answer to the questions above will help) I believe I've now finished all the necessary edits to the API and manifest pages. I will double check tomorrow, but should you have the chance to take a look and provide feedback that would be appreciated.

:+1: I'll give it a look to the related MDN pages asap

rebloor:
@rpl Thanks for the feedback I've made some further updates and eliminated the separate APIScripts page. I do have a couple of questions:

* Is the example on [userScripts.onBeforeScript](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/onBeforeScript) an example of the script you would register with user_scripts, or something else?
* Presumably [userScripts.RegisteredUserScript](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/RegisteredUserScript) contains the same properties as provided in [userScripts.register](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/user_scripts) and we should add documentation?

rpl:
> @rpl Thanks for the feedback I've made some further updates and eliminated the separate APIScripts page. I do have a couple of questions:
> 
>     * Is the example on [userScripts.onBeforeScript](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/onBeforeScript) an example of the script you would register with user_scripts, or something else?

@rebloor The example in that page is the example of an API script (registered in the manifest using the `user_scripts.api_script` property) that is providing some custom APIs to the user scripts registered
by the same extension.

(in general the userScripts doesn't have access to any of the WebExtensions API, the `api_script` on the contrary is allowed to use any of the WebExtensions APIs available to a content script, including the special `userScripts.onBeforeScript` which the api_script should use to access the sandboxes created for the matched userScripts).
 
>     * Presumably [userScripts.RegisteredUserScript](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/RegisteredUserScript) contains the same properties as provided in [userScripts.register](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/user_scripts) and we should add documentation?

RegisteredUserScript is the API schema type of the object that the promise returned by `userScripts.register` resolves to, it currently only provides a single `unregister` method,
see schema definition here: https://searchfox.org/mozilla-central/rev/053826b10f838f77c27507e5efecc96e34718541/toolkit/components/extensions/schemas/user_scripts.json#90-103

rebloor:
@rpl Thanks for your replies. Just to confirm, [userScripts.RegisteredUserScript](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/RegisteredUserScript) only contains the `unregister` method, it has no properties?

I believe, therefore, that my changes to the API documentation are complete. Would you have the opportunity, if you haven't done so already, to do a full review? @chrisdavidmills should I get anyone else to review?

I will now look at updating the example/example description.

rpl:
> @rpl Thanks for your replies. Just to confirm, [userScripts.RegisteredUserScript](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/RegisteredUserScript) only contains the `unregister` method, it has no properties?

yes, I confirm that `RegisterUserScript` doesn't currently contain any other properties besides the `unregister` method.

> I believe, therefore, that my changes to the API documentation are complete. Would you have the opportunity, if you haven't done so already, to do a full review?

Sure, I'm going to do a full review on my side.

rebloor:
@rpl have you been able to do that "full review" of the MDN content?

Jeffhurd92:
Girls1

Jeffhurd92:
``

wbamberg:
Closed in favour of https://github.com/mdn/webextensions-examples/pull/438.

PR #428, opened 2019-10-16
replaced cat gif URL as it was outdated

pastparty:
## What's in this pull request?
A changed URL as the former one points to an oudated website.


rebloor:
The extension example now uses https://giphy.com/explore/cat

PR #438, opened 2020-01-29
New user script example

rebloor:
This PR is the content from [PR 426](https://github.com/mdn/webextensions-examples/pull/426), originally submitted by Irene, addressing the feedback on the readme content.


rebloor:
@wbamberg  I'm wondering whether this should replace the existing user script example, the one based on using content scripts?

user-script-register-v2/README.md
```diff
@@ -0,0 +1,19 @@
+# User script
+
+This extension demonstrates the [`browser.userScripts.register()`](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/RegisteredUserScript) API. This API enables an extension to register URL-matching scripts at runtime, scripts that run in isolated sandboxes.
```
wbamberg: I guess the link here should be to https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/register ? Also I don't think the second sentence is helpful here, probably better just to let people read the docs.

user-script-register-v2/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+The extension includes an API script (`customUserScriptAPIs.js`) that provides stubs that enable a registered script to make use of `browser.storage.local`.
```
wbamberg: "API script" could helpfully link to https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/user_scripts. And these are not stubs: stubs do nothing. These functions wrap or expose the `storage.local` functions.

user-script-register-v2/README.md
```diff
@@ -0,0 +1,19 @@
...
+The extension includes an API script (`customUserScriptAPIs.js`) that provides stubs that enable a registered script to make use of `browser.storage.local`. 
+
+To enable a user script to be specified and registered, this extension includes a sidebar action. The sidebar enables you to define all the properties associated with controlling the execution of a registered script. The example includes values for:
+
+* the host pattern `*://*.org/*`, which loads the script into any HTTP or HTTPS pages on a `.org` domain.
+* script code that replaces the content of the pattern matched page with the message "This page has been eaten". The script also uses the API script stubs to save and recall the URL of each page "eaten". Information on the last and current "eaten" page is then included in the "eaten" message.
+
+All other properties are empty or given their default value.
+
+Clicking "Register script" register the script by calling `browser.userScripts.register()`.
```
wbamberg: register -> registers

user-script-register-v2/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+You can only have one script registered at any time: registering a new script unregisters the existing script. The extension does this by keeping a reference to the `RegisteredUserScript` object returned from `browser.userScripts.register()`: this object provides the `unregister()` method.
```
wbamberg: Might be worth clarifying: *In this example*, you can only have..." - since in the API itself you can of course have > 1.

user-script-register-v2/background.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+var registered = null;
```
wbamberg: Could use `let` here.

user-script-register-v2/background.js
```diff
@@ -0,0 +1,36 @@
...
+
+async function registerScript(message) {
+  let {
```
wbamberg: Could be `const`.

user-script-register-v2/manifest.json
```diff
@@ -0,0 +1,40 @@
+{
+
+  "manifest_version": 2,
+  "name": "User script",
+  "version": "2.0",
+
+  "applications": {
+    "gecko": {
+      "id": "user-script-example@mozilla.org",
```
wbamberg: This has the same ID as "user-script", so if we do want to keep both, it should be changed.

user-script-register-v2/popup/user-script.html
```diff
@@ -0,0 +1,63 @@
...
+
+<html>
+  <head>
+    <meta charset="utf-8">
+    <link rel="stylesheet" href="user-script.css"/>
+  </head>
+
+<body>
+
+  <div id="outer-wrapper">
```
wbamberg: I think you don't need this. It seems to be used only in the CSS:

```css
#outer-wrapper {
  padding: 0.5em;
}
```

but you can just do:

```css
body {
  padding: 0.5em;
}
```

user-script-register-v2/popup/user-script.html
```diff
@@ -0,0 +1,63 @@
...
+    <label for="includeGlobs">Include Globs</label>
+    <input id="includeGlobs" type="text">
+
+    <label for="excludeGlobs">Exclude Globs</label>
+    <input id="excludeGlobs" type="text">
+
+    <label for="matchAboutBlank">Match About Blank</label>
+    <select id="matchAboutBlank">
+      <option selected value="false">false</option>
+      <option value="true">true</option>
```
wbamberg: It would be better to use a checkbox for this.

user-script-register-v2/popup/user-script.html
```diff
@@ -0,0 +1,63 @@
...
+    </select>
+    <br/>
+
+    <label for="matchAboutBlank">All Frames</label>
+    <select id="allFrames">
+      <option selected value="false">false</option>
+      <option value="true">true</option>
```
wbamberg: Again it would be better to use a checkbox.

rebloor:
Hi @wbamberg  just checking in to make sure you'd seen my commit with changes to address your feedback.

content-script-register/README.md
```diff
@@ -1,24 +1,26 @@
-# User script
+# Content script registration
 
 This extension demonstrates the `browser.contentScripts.register()` API, which enables an extension to register URL-matching content scripts at runtime.
 
+If you want functionality to load user scripts, scripts that can be provided access to extension APIs not available to content scripts, please refer to the [user script registration](https://github.com/mdn/webextensions-examples/tree/master/user-script-register) example.
```
wbamberg: I think this is misleading:

> If you want functionality to load user scripts, **scripts that can be provided access to extension APIs not available to content scripts**

This suggests to me that the key difference between content scripts and user scripts is that an extension can give extra powers to user scripts - specifically, "access to extension APIs not available to content scripts".

From my reading of https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/user_scripts, it's **not the case** that user scripts can give this kind of extra access, because the API script, that can essentially provide extra APIs to user scripts:

> has access to the same subset of WebExtension APIs usually available in a content script.

It might be better to say something like:

> The contentScripts.register() API is intended to enable an extension to register at runtime scripts that are packaged with the extension itself. If you want to register third-party user scripts, you should instead use the [`userScripts`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts) API. Please refer to [user-script-register](https://github.com/mdn/webextensions-examples/tree/master/user-script-register) for an example of that API.

user-script-register/README.md
```diff
@@ -0,0 +1,19 @@
...
+This extension demonstrates the [`browser.userScripts.register()`](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/Register) API.  
+
+The extension includes an [API script](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/user_scripts) (`customUserScriptAPIs.js`) that registers a script that enables user scripts to make use of `browser.storage.local`. 
+
+To enable a user script to be specified and registered, this extension includes a sidebar action. The sidebar enables you to define the following properties that control the execution of a registered script, with default values provided:
+
+* the host pattern `*://*.org/*`, which loads the script into any HTTP or HTTPS pages on a `.org` domain.
+* script code that replaces the content of the pattern matched page with the message "This page has been eaten". The script also uses the API script stubs to save and recall the URL of each page "eaten". Information on the last and current "eaten" page is then included in the "eaten" message.
+
+All other properties are take their default value.
```
wbamberg:
```suggestion
All other properties are given their default value.
```

user-script-register/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Clicking "Register script" registers the script by calling `browser.userScripts.register()`.
+
+In this example, only one script can be registered at a time: registering a new script unregisters the active script. The extension does this by keeping a reference to the `RegisteredUserScript` object returned from `browser.userScripts.register()`: this object provides the `unregister()` method.
+
+To try the extension, click "Register script" with the defaults and load http://example.org/ or 
+https://www.mozilla.org/. Then change the pattern or the code and reload's these or related pages.
```
wbamberg:
```suggestion
https://www.mozilla.org/. Then change the pattern or the code and reload these or related pages.
```

user-script-register/popup/user-script.js
```diff
@@ -0,0 +1,95 @@
...
+  lastErrorEl.textContent = params.lastError || "";
+}
+
+function stringToArray(value) {
+  const res = value.split(",").map(el => el.trim()).filter(el => el !== "");
+
+  return res.length > 0 ? res : null;
+}
+
+function stringToBool(value) {
```
wbamberg: I think this is now unused and can be deleted?

content-script-register/README.md
rebloor: Good point, amended

user-script-register/README.md
rebloor: Perhaps "use" would be more accurate, as the code doesn't do anything to give the other  properties their default values

user-script-register/README.md
rebloor: Fixed

user-script-register/popup/user-script.js
rebloor: Thanks for spotting that one, removed

rpl:
@rebloor this review is in my review queue, but I haven't been able to reach it yet. I'll try to come back to this asap (my apologies for not being able to give it a review round yet)

user-script-register/customUserScriptAPIs.js
```diff
@@ -0,0 +1,19 @@
...
+    async GM_getValue(name) {
+      const res = await browser.storage.local.get(name);
+      console.log("GM_getValue", {name, res, scriptMetadata});
+      return res[name];
+    },
+    GM_setValue(name, value) {
+      console.log("GM_setValue", {name, value, scriptMetadata});
+      return browser.storage.local.set({[name]: value});
+    },
+  });
```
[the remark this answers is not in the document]
rebloor: @rpl  Sorry for the slow response. Looks like an excellent idea and I have implemented it in the latest commit. While testing your suggested changes l found that in:
```
    async GM_getValue(name) {
      const res = await browser.storage.local.get(`${id}:${name}`);
      console.log("GM_getValue", {name, res, scriptMetadata});
      return res[name];
    },
```
the `return` didn't work and I modified it to `return res[`${id}:${name}`];` which feels inelegant, is there a better solution?

FYI @wbamberg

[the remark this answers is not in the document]
wbamberg: > is there a better solution?

I wasn't sure if you were asking me too, but I might do something like:

```js
browser.userScripts.onBeforeScript.addListener(script => {

  const scriptMetadata = script.metadata;
  const id = scriptMetadata.userScriptID;

  function getScopedName(name) {
    return `${id}:${name}`;
  }

  script.defineGlobals({
    async GM_getValue(name) {
      const scopedName = getScopedName(name);
      const res = await browser.storage.local.get(scopedName);
      console.log("GM_getValue", {id, name, res, scriptMetadata});
      return res[scopedName];
    },
    GM_setValue(name, value) {
      console.log("GM_setValue", {id, name, value, scriptMetadata});
      return browser.storage.local.set({[getScopedName(name)]: value});
    },
  });

  console.log("custom userScripts APIs defined");
});

console.log("apiScript executed and userScripts.onBeforeScript listener subscribed");
```

...i.e. have a separate function that glues `id` and `name`, to make it clear that both get and set have to do the same operation.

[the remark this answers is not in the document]
rebloor: > I wasn't sure if you were asking me too
Really anyone who might be listening!
Thanks, @wbamberg I've implemented your suggestion.  If everything else looks OK, would you be happy to approve or would you prefer to wait for @rpl to respond?

wbamberg:
Thanks @rebloor !

PR #445, opened 2020-03-04
quicknote chrome compatibility

DaWe35:
I want to make addons for every browser, not just chrome.
`var browser = browser || chrome;` helped me out, but there was something what did not worked in Chrome. After some research, I found [this page](https://developer.chrome.com/apps/storage).

```
chrome.storage.local.get(['key'], function(result) {
  console.log('Value currently is ' + result.key);
});
```
It's compatible with Firefox and Chrome also, but I lost the `onerror` callback. How can I get back it without breaking the compatibility?

### DIFF
--- a/proxy-blocker/options/options.js
+++ b/proxy-blocker/options/options.js
@@ -14,12 +14,10 @@ function updateUI(restoredSettings) {
   blockedHostsTextArea.value = restoredSettings.blockedHosts.join("\n");
 }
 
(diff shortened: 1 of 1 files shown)


caitmuenster:
Hi @DaWe35, apologies for the late reply. Your question will get more visibility if you post it on the [add-ons community forum](https://discourse.mozilla.org/c/add-ons/development/). :)

PR #457, opened 2020-08-19
Use async/await in options.js

Flimm:


lucalves (approved):

rebloor:
@rpl I'm in the process of tidying up some of these old PRs. I've exercised the example, and it still works as expected. The code looks OK to me, but could I get a second opinion?

rpl (approved):
> @rpl I'm in the process of tidying up some of these old PRs. I've exercised the example, and it still works as expected. The code looks OK to me, but could I get a second opinion?

lgtm too

Flimm:
Thank you for the merge!

Issue #364, opened 2018-07-21
browser.webRequest.filterResponseData ondata  error


admimistrator:
https://github.com/mdn/webextensions-examples/tree/master/http-response
code :
```
function listener(details) {
  var filter = browser.webRequest.filterResponseData(details.requestId);
  var decoder = new TextDecoder("utf-8");
  var encoder = new TextEncoder();

  filter.ondata = event => {
	
    var str = decoder.decode(event.data, {stream: true});
   console.log( str.length, str);
    filter.write(encoder.encode(str));
    filter.disconnect();
  }

  return {};
}

browser.webRequest.onBeforeRequest.addListener(
  listener,
  {urls: ["<all_urls>"], types: ["xmlhttprequest"]},
  ["blocking"]
);
```
console.log( str.length, str);
sometimes, the  str  is fragmentary .
for example,like this pic:
![qq 20180722004155](https://user-images.githubusercontent.com/9912143/43038229-2bcb1ebc-8d48-11e8-812a-4102b6e50e92.png)

"alig" it's not the end of response ,but it's end...

the real response is :
![qq 20180722004826](https://user-images.githubusercontent.com/9912143/43038284-2da4345c-8d49-11e8-8eeb-7ec0c0dd3cc6.png)

dyeray:
The example is pretty lousy. This API is streamed, so filter.ondata is called on chunks of the response, not the full response, but if you call filter.disconnect() you stop the process (only printing the first chunk). This article has a bit better examples: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/StreamFilter/ondata.

seanlinsley:
@dyeray thanks for that link. Coming here from https://stackoverflow.com/questions/46338823, I couldn't figure out why I was only getting the first chunk of data, and if I removed the `disconnect()`, the response never made it to its original destination.

seanlinsley:
:heavy_check_mark: added a note to StackOverflow and the [top-level MDN docs](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/filterResponseData)

rebloor:
No activity within the last three years, closing.

PR #460, opened 2020-11-05
Make http response example more robust

wingman-jr-addon:
I've been using WebRequest filtering for quite some time now in one of my plugins, and while it's easy to get started with toy examples, I've found that there is some significant complexity that must be addressed when trying to do any sort of text manipulation in a bit more robust manner. (Some of these deficiencies are discussed in #364 , partially addressed in the MDN docs ) I thought extending this extension would be the right place to start as I would have wished to know these details when I started my plugin.

This extension now demonstrates three important things:
 - The accumulation of data through multiple calls to `.ondata`
 - The decoding of binary data to text in a streaming fashion. (Not necessarily performant, but should be correct.)
 - Text decoding that tries to respect the page's reported encoding via `Content-Type`.

As an observation, browser support for two things would clean this up tremendously:

- Exposing the detected character encoding on the `webRequest` details (at least insofar as it is detectable based on headers) would avoid manual, poor detection logic like that I have here.
- Adding support for different encodings for `TextEncoder` so that the original character encoding can be preserved rather than always converting to UTF-8.

At any rate, let me know what you think! I hope this will help provide a bit stronger starting point for others using `webRequest`.

### DIFF
--- a/http-response/background.js
+++ b/http-response/background.js
@@ -1,22 +1,153 @@
 function listener(details) {
+  // The received data is a stream of bytes. In order to do text-based
+  // modifications, it is necessary to decode the bytes into a string
+  // using the proper character encoding, do any modifications, then
+  // encode back into a stream of bytes.
+  // Historically, detecting character encoding has been a tricky task
+  // taken on by the browser. Here, a simplified approach is taken
+  // and the complexity is hidden in a helper method.
+  let decoder, encoder;
+  [decoder, encoder] = detectCharsetAndSetupDecoderEncoder(details);
   let filter = browser.webRequest.filterResponseData(details.requestId);
-  let decoder = new TextDecoder("utf-8");
-  let encoder = new TextEncoder();
-
-  filter.ondata = event => {
-    let str = decoder.decode(event.data, {stream: true});
-    // Just change any instance of Example in the HTTP response
-    // to WebExtension Example.
-    str = str.replace(/Example/g, 'WebExtension Example');
-    filter.write(encoder.encode(str));
-    filter.disconnect();
+  let fullStr = '';
+  
+  filter.ondata = e => {
+    // Note that the event's data may break in the middle of an encoded
+    // character - the stream parameter is critical for success as this
+    // method gets called multiple times.
+    let str = decoder.decode(e.data, {stream: true});
+    fullStr += str;
+  }
+  
+  filter.onstop = async e => {
+    // Just change any instance of Example or Test in the HTTP response
+    // to WebExtension Example or WebExtension Test.
+    let mutatedStr = fullStr.replace(/Example/g, 'WebExtension Example');
+    mutatedStr = mutatedStr.replace(/Test/g, 'WebExtension Test');
+    filter.write(encoder.encode(mutatedStr));
+    filter.close();
   }
 
-  return {};
+  filter.onerror = e => {
+    try {
+        filter.close();
+        console.log('Filter error: '+e+', '+ex);
+    } catch(ex) {
+        console.log('Filter error while closing: '+e+', '+ex);
+    }
+  }
+
+  // Because details response headers have been mutated, return it
+  return details;
 }
 
-browser.webRequest.onBeforeRequest.addListener(
+browser.webRequest.onHeadersReceived.addListener(
   listener,
-  {urls: ["https://example.com/*"], types: ["main_frame"]},
-  ["blocking"]
+  {
+    urls: ["https://example.com/*", "https://www.w3.org/*"], // Include W3 for testing charset detection.
+    types: ["main_frame"]
+  },
+  ["blocking","responseHeaders"]
 );
+
+// This helper method does a few things regarding character encoding:
+// 1) Detects the charset for the TextDecoder so that bytes are properly turned into strings
+// 2) Ensures the output Content-Type is UTF-8 because that is what TextEncoder supports
+// 3) Returns the decoder/encoder pair
+function detectCharsetAndSetupDecoderEncoder(details) {
+  let contentType = '';
+  let headerIndex = -1;
+  for(let i=0; i<details.responseHeaders.length; i++) {
+      let header = details.responseHeaders[i];
+      if(header.name.toLowerCase() == "content-type") {
+          contentType = header.value.toLowerCase();
+          headerIndex = i;
+          break;
+      }
+  }
+  if (headerIndex == -1) {
+    console.log('No Content-Type header detected for '+details.url+', adding one.');
+    headerIndex = details.responseHeaders.length;
+    contentType = 'text/html';
+    details.responseHeaders.push(
+      {
+        "name": "Content-Type",
+        "value":"text/html"
+      }
+    );
+  }
+
+  let baseType;
+  if(contentType.trim().startsWith('text/html')) {
+    baseType = 'text/html';
+    console.log('Detected base type was '+baseType);
+  } else if(contentType.trim().startsWith('application/xhtml+xml')) {
+    baseType = 'application/xhtml+xml';
+    console.log('Detected base type was '+baseType);
+  } else {
+    baseType = 'text/html';
+    console.log('The Content-Type was '+contentType+', not text/html or application/xhtml+xml - results might be strange.');
+  }
+
+  // It is important to detect the charset to correctly initialize TextDecoder or
+  // else we run into garbage output sometimes.
+  // However, TextEncoder does NOT support other than 'utf-8', so it is necessary
+  // to change the Content-Type on the header to UTF-8
+  // If modifying this block of code, ensure that the tests at
+  // https://www.w3.org/2006/11/mwbp-tests/index.xhtml
+  // all pass - current implementation only fails on #9 but this detection ensures
+  // tests #3,4,5, and 8 pass.
+  let decodingCharset = 'utf-8';
+  let detectedCharset = detectCharset(contentType);
+
+  if(detectedCharset !== undefined) {
+      decodingCharset = detectedCharset;
+      console.log('Detected charset was ' + decodingCharset + ' for ' + details.url);
+  }
+  details.responseHeaders[headerIndex].value = baseType+';charset=utf-8';
+
+  let decoder = new TextDecoder(decodingCharset);
+  let encoder = new TextEncoder(); //Encoder does not support non-UTF-8 charsets so this is always utf-8.
+
+  return [decoder,encoder];
+}
+
+// Detect the charset from Content-Type
+function detectCharset(contentType) {
+  /*
+  From https://tools.ietf.org/html/rfc7231#section-3.1.1.5:
+
+  A parameter value that matches the token production can be
+  transmitted either as a token or within a quoted-string.  The quoted
+  and unquoted values are equivalent.  For example, the following
+  examples are all equivalent, but the first is preferred for
+  consistency:
+
+  text/html;charset=utf-8
+  text/html;charset=UTF-8
+  Text/HTML;Charset="utf-8"
+  text/html; charset="utf-8"
+
+  Internet media types ought to be registered with IANA according to
+  the procedures defined in [BCP13].
+
+  Note: Unlike some similar constructs in other header fields, media
+  type parameters do not allow whitespace (even "bad" whitespace)
+  around the "=" character.
+
+  ...
+
+  And regarding application/xhtml+xml, from https://tools.ietf.org/html/rfc3236#section-2
+  and the referenced links, it can be seen that charset is handled the same way with
+  respect to Content-Type.
+  */
+
+  let charsetMarker = "charset="; // Spaces *shouldn't* matter
+  let foundIndex = contentType.indexOf(charsetMarker);
+  if (foundIndex == -1) {
+      return undefined;
+  }
+  let charsetMaybeQuoted = contentType.substr(foundIndex+charsetMarker.length).trim();
+  let charset = charsetMaybeQuoted.replace(/\"/g, '');
+  return charset;
+}
--- a/http-response/background.js
+++ b/http-response/background.js
@@ -20,6 +20,7 @@ function listener(details) {
   }
   
   filter.onstop = async e => {
+    fullStr += decoder.decode(); //Flush the buffer
     // Just change any instance of Example or Test in the HTTP response
     // to WebExtension Example or WebExtension Test.
     let mutatedStr = fullStr.replace(/Example/g, 'WebExtension Example');
--- a/http-response/README.md
+++ b/http-response/README.md
@@ -3,12 +3,12 @@
 ## What it does
 
 Listens to HTTP Responses from example.com and w3.org and changes "Example" to "WebExtension Example" and
-"Test" to "WebExtension Test" in the web pages contents. 
+"Test" to "WebExtension Test" in the web pages' contents. 
 
 ## What it shows
 
 A real-world example of WebRequest that shows three important details not always found in beginning examples:
- - The accumulation of data through multiple calls to .ondata
+ - The accumulation of data through multiple calls to `.ondata`
  - The decoding of binary data to text in a streaming fashion.
  - Text decoding that tries to respect the page's reported encoding via Content-Type.
 
--- a/http-response/README.md
+++ b/http-response/README.md
@@ -2,10 +2,19 @@
 
 ## What it does
 
-Listens to HTTP Responses from example.com and changes the body of the response as it comes through. So that the word "Example" on https://example.com becomes "WebExtension Example".
+Listens to HTTP Responses from example.com and w3.org and changes "Example" to "WebExtension Example" and
+"Test" to "WebExtension Test" in the web pages contents. 
 
 ## What it shows
 
-How to use the response parser on bytes.
+A real-world example of WebRequest that shows three important details not always found in beginning examples:
+ - The accumulation of data through multiple calls to .ondata
+ - The decoding of binary data to text in a streaming fashion.
+ - Text decoding that tries to respect the page's reported encoding via Content-Type.
+
+The domain w3.org is included in the list of domains to allow for testing against [this suite of standardized tests](https://www.w3.org/2006/11/mwbp-tests/index.xhtml)
+regarding text encoding. Tests #1-8 pass, test #9 currently fails.
+
+## Credits
 
 Icon is from: https://www.iconfinder.com/icons/763339/draw_edit_editor_pen_pencil_tool_write_icon#size=128
--- a/http-response/manifest.json
+++ b/http-response/manifest.json
@@ -10,7 +10,7 @@
   },
 
   "permissions": [
-    "webRequest", "webRequestBlocking", "https://example.com/*"
+    "webRequest", "webRequestBlocking", "https://example.com/*", "https://www.w3.org/*"
   ],
 
   "background": {


wingman-jr-addon:
@caitmuenster Here is the PR that I was curious about; you can see [the question I posted on Discourse as well](https://discourse.mozilla.org/t/webrequest-charset-detection/69910) if you're curious. Thanks!

caitmuenster:
Thanks @wingman-jr-addon! Adding this to @Rob--W's queue for review. :)

http-response/background.js
```diff
@@ -1,22 +1,154 @@
 function listener(details) {
+  // The received data is a stream of bytes. In order to do text-based
+  // modifications, it is necessary to decode the bytes into a string
+  // using the proper character encoding, do any modifications, then
+  // encode back into a stream of bytes.
+  // Historically, detecting character encoding has been a tricky task
+  // taken on by the browser. Here, a simplified approach is taken
+  // and the complexity is hidden in a helper method.
+  let decoder, encoder;
+  [decoder, encoder] = detectCharsetAndSetupDecoderEncoder(details);
```
Rob--W, line 10: The logic is complicated but still not a full solution to the problem.

For simplicity, I recommend to check whether the response is (X)HTML (to ignore responses such as images, media, downloads), and only proceed if needed. Then the MIME/charset detection logic can greatly be simplified (see also comment below about charset detection).

You should also check that the HTTP status code is OK before attaching a stream filter.

http-response/background.js
```diff
@@ -1,22 +1,154 @@
...
+    let str = decoder.decode(e.data, {stream: true});
+    fullStr += str;
+  }
+  
+  filter.onstop = async e => {
+    fullStr += decoder.decode(); //Flush the buffer
+    // Just change any instance of Example or Test in the HTTP response
+    // to WebExtension Example or WebExtension Test.
+    let mutatedStr = fullStr.replace(/Example/g, 'WebExtension Example');
+    mutatedStr = mutatedStr.replace(/Test/g, 'WebExtension Test');
```
Rob--W, line 27: Remove the second one. It is not necessary for the example.

http-response/background.js
```diff
@@ -1,22 +1,154 @@
...
-  filter.ondata = event => {
-    let str = decoder.decode(event.data, {stream: true});
-    // Just change any instance of Example in the HTTP response
-    // to WebExtension Example.
-    str = str.replace(/Example/g, 'WebExtension Example');
-    filter.write(encoder.encode(str));
-    filter.disconnect();
+  filter.onerror = e => {
+    try {
+        filter.close();
```
Rob--W, line 34: Why do you want to explicitly handle errors here?

`filter.onerror` could occur when the response is redirected (when you've failed to check for status codes, as you did before),
and also repeatedly when an out-of-memory issue occurs - https://bugzilla.mozilla.org/show_bug.cgi?id=1404381

http-response/background.js
```diff
@@ -1,22 +1,154 @@
...
-  let decoder = new TextDecoder("utf-8");
-  let encoder = new TextEncoder();
+  let fullStr = '';
+  
+  filter.ondata = e => {
+    // Note that the event's data may break in the middle of an encoded
+    // character - the stream parameter is critical for success as this
+    // method gets called multiple times.
+    let str = decoder.decode(e.data, {stream: true});
+    fullStr += str;
```
Rob--W, line 19: Buffering everything before flushing is simple, but not user friendly.

Instead of buffering everything at once, you can consider flushing all but some penultimate bytes/characters. After all, if the string to replace is at most 7 characters, then you can safely flush everything except the last 6 characters, to handle this case:

```
[prefix this can be flushed]Exampl
```

http-response/background.js
```diff
@@ -1,22 +1,154 @@
...
+    types: ["main_frame"]
+  },
+  ["blocking","responseHeaders"]
 );
+
+// This helper method does a few things regarding character encoding:
+// 1) Detects the charset for the TextDecoder so that bytes are properly turned into strings
+// 2) Ensures the output Content-Type is UTF-8 because that is what TextEncoder supports
+// 3) Returns the decoder/encoder pair
+function detectCharsetAndSetupDecoderEncoder(details) {
```
Rob--W, line 58: "detect and set up decoder/encoder" suggests that it just does detection and creating new instances of the decoder/encoder.

The method however also mutates the response headers.

http-response/background.js
```diff
@@ -1,22 +1,154 @@
...
+  details.responseHeaders[headerIndex].value = baseType+';charset=utf-8';
+
+  let decoder = new TextDecoder(decodingCharset);
+  let encoder = new TextEncoder(); //Encoder does not support non-UTF-8 charsets so this is always utf-8.
+
+  return [decoder,encoder];
+}
+
+// Detect the charset from Content-Type
+function detectCharset(contentType) {
```
Rob--W, line 117: The charset detection logic as implemented below is incorrect (e.g. when multiple `charset` parameters are set, the last one should be used; when multiple MIME types are set, the previously set charset should be forgotten). Let's keep the example simple, however, and make it more obvious that the implementation is incomplete.

Here is an example of a more complete implementation that also covers some edge cases: https://github.com/Rob--W/open-in-browser/commit/a6b926ea9522b35298632e5e6a2c89ddb456c5d9

As you can see, it's quite involved - do not put the code in the PR. You can link to it if you wish though.

http-response/background.js
```diff
@@ -1,22 +1,154 @@
 function listener(details) {
+  // The received data is a stream of bytes. In order to do text-based
+  // modifications, it is necessary to decode the bytes into a string
+  // using the proper character encoding, do any modifications, then
+  // encode back into a stream of bytes.
+  // Historically, detecting character encoding has been a tricky task
+  // taken on by the browser. Here, a simplified approach is taken
+  // and the complexity is hidden in a helper method.
+  let decoder, encoder;
```
Rob--W, line 9: Keep the `TextDecoder` and `TextEncoder` constructors here as before to improve readability. You can declare a `charset` variable if you want to use something else instead of UTF-8.

Rob--W (changes requested):
In an example we need to strike a balance between completeness of code and simplicity of code. Currently it is not simple, nor complete - see my comments below.

wingman-jr-addon:
Thanks @Rob--W  for taking the time to look at this and provide helpful feedback, and thanks @caitmuenster for getting this going. [I haven't gotten much feedback from the community](https://discourse.mozilla.org/t/webrequest-charset-detection/69910) on this aspect of using WebRequest and yet I think it is so vital to any addon trying to filter text. I'll take a look at your comments; I think you might have some better ideas on how to solve some of the problems I'm facing but it may take a round or two of feedback until I get into the right spot. Thanks!

http-response/background.js
wingman-jr-addon: Yes, I don't like it either. I believe that the real issue here is that TextEncoder does not support anything other than UTF-8, so we **must** force the content type on the output, which is awkward. As noted in the PR description, the code would benefit from having TextEncoder support more types to avoid this hack. Can you think of a way to preserve the original content encoding without the TextEncoder polyfill? I'd rather bypass the mutation altogether because it seems a bit treacherous.

http-response/background.js
wingman-jr-addon: Well, except that I need to potentially modify the Content-Type output to be UTF-8 if the input type is not already; that would drive having something like `detectCharset` and `forceCharset` as separate calls. Would you find that organization to be preferable to the problematic `detectCharsetAndSetupDecoderEncoder` above?

http-response/background.js
wingman-jr-addon: It is necessary for the example. I added it because the W3C test suite pages do not have the word "Example" but do have "Test" so it shows that the plugin is mutating the pages when viewing the test suite - I think that's a critical aspect to preserve. However, in the name of simplicity, I could remove the "Example"/example.com from the addon; I was trying to keep a nod to the original behavior but perhaps it just feels kludgy now.

http-response/background.js
wingman-jr-addon: That's a good point. This is an artifact of the code I was porting this from (where I needed to do a regex search of unknown length) and can be improved. I think the best solution for this would be a streaming regex transformer - I don't suppose something along those lines exists without external library support? That would make my day.  Otherwise, I will change it as requested.

http-response/background.js
wingman-jr-addon: Good question - this part of the error handling isn't particularly relevant to the heart of the example either.

- [ ] Remove explicit error handling

http-response/background.js
wingman-jr-addon: The note about the complex logic is discussed in the other relevant comments below, but duly noted.

I agree it would be better to move the logic for detecting the supported MIME types as a pre-check here.
- [ ] Pre-check MIME types

Also agreed on the HTTP status code check.

- [ ] Add HTTP status code check

http-response/background.js
wingman-jr-addon: There's wisdom here; thanks for the link.
I've been thinking about your comment about an obviously incomplete implementation. While the idea of a complete solution here is appealing, the simplified solution will likely cover 99% of the cases and the comment will make the careful reader aware of the problem. Lack of awareness of this problem - albeit obvious in retrospect - is why I'm wanting to contribute back a more robust example in the first place; a user reported a bug to me in a review of my addon. I'd like others to avoid the same fate, as their users may not be as considerate as mine was.

- [ ] Make incompleteness obvious

Given both the complexity of determining the charset as well as the need for it to be accurate in order to filter text properly, do you think the browser-predicted charset would ever be a candidate to expose on the `details` object? It just seems like right now it would be incredibly difficult to create a truly correct implementation - do you know of any extant addons that are actually correct in this regard?

http-response/background.js
Rob--W: If you'd like to replace both, you could do `/Example|Test/` to demonstrate how one can replace multiple words (and if you wish to prepend "WebExtension", use `'WebExtension $&'`. If you don't know about `$&`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter

http-response/background.js
Rob--W: You don't need an external library for this, the logic can be written in a concise way. See the brief explanation below for the algorithm. It is definitely more involved than buffering all and flushing at once though. If the example becomes very unwieldy if this were to be implemented, then buffering may not be that bad of an idea.

Instead of using `replace` with a regexp and a string, you can use the `.exec` method of a regular expression with the `g` flag to find matches. Flush the prefix, append the replacement, and continue (if you are cutting the string, reset `lastIndex` to `0` to resume from the start, if you aren't cutting the string you can just call `.exec` again and it will continue searching from position `lastIndex` in the string).

When you don't find a match, and know that the regular expression has a maximum number of matches (such as `8` in case of `/Example|Test/g`), then cut the remaining string in two parts: Keep the last 7 characters (=max max length minus one) for the next time and flush the rest (this is what I explained in my previous comment).

Note: when your regular expression does not have a unique match (e.g. `/A{1,3}/g` or `[0-9]+`), then you need extra logic: to make sure that as much is matched as possible, when the regular expression matches the end of the string collected so far, you should not flush and keep the string for the next iteration, in case the regexp is going to match more.

http-response/background.js
Rob--W: For search keywords that are compatible with the used encodings (e.g. letters from the latin alphabet have the same bytes in utf-8, iso-8859-2, ..), you don't need to worry about using TextEncoder. For fixed-length search terms, you could convert the string to bytes and search and replace on typed arrays instead of strings. This may be even more performant than converting to strings and back.

But for the example, I think that it is fine to convert to UTF-8. First, it shows the significance of character encodings.
For those who are not that well versed with character encodings, it is a safe default to choose, because utf-8 is very common.

http-response/background.js
Rob--W: > Given both the complexity of determining the charset as well as the need for it to be accurate in order to filter text properly, do you think the browser-predicted charset would ever be a candidate to expose on the `details` object?

What a good idea. I have thought of it before and posted a feature request at https://bugzilla.mozilla.org/show_bug.cgi?id=1425479
Unfortunately the feature request was denied. If you post a new bug that shows use cases and why alternatives aren't a good fit, the feature request can be re-evaluated.

>  It just seems like right now it would be incredibly difficult to create a truly correct implementation - do you know of any extant addons that are actually correct in this regard?

The Open in Browser add-on that I linked before was authored by me, and modelled after Firefox's implementation (which includes non-standard behavior).

http-response/background.js
Rob--W: In the simplest case where the Content-Type needs to be an exact match for something that looks like `text/html; charset=utf-8", then you don't need to change the response header. But even if you do change the header, then you can still improve the readability by keeping the main logic in the listener, and using small helper functions to avoid too large function bodies. For example:

```javascript
function listener(details) {
  let {responseHeaders} = details;

  let ctHeader = responseHeaders.find(h => h.name.toLowerCase() == "content-type");
  // TODO: Verify that ctHeader is supported.
  // If Content-Type header not set, the browser is going to do content-sniffing,
  // and we should also return to avoid trouble (e.g. breaking downloads, PDFs, videos, ...).
  if (content type is not supported) {
    return;
  }

  let charset = ...
  let decoder = new TextDecoder(charset);
  let encoder = new TextEncoder();
  etc...
}
```

http-response/background.js
wingman-jr-addon: Thanks - I think your first comment made me realize that I had a conflict of "make everyone happy" or "have a better focused example" and so I'm choosing the latter and keeping it simple. As venerable as example.com is, it gets enough press elsewhere.

http-response/background.js
wingman-jr-addon: I opted to go for the simpler static string streaming replacement, and added comments around its limits.

http-response/background.js
wingman-jr-addon: I took another run at the code. I was able to clean it up in a few places as per your comments - keep the header object reference rather than index, return earlier/don't guess, etc. Still, a full *half* of the code is dedicated to character encoding alone; I can't say that the code is elegant.

wingman-jr-addon:
@Rob--W I've taken another pass at the code and have closed out the open tasks I set up on the review. Ultimately, I still dislike the code, but I think it's an improvement and thanks for your help. Let me know if it passes muster.

I'd also like to promote to the top level the conversation we had going in the sub-thread about e.g. exposing the detected MIME type. We were saying:

> >    Given both the complexity of determining the charset as well as the need for it to be accurate in order to filter text properly, do you think the browser-predicted charset would ever be a candidate to expose on the details object?

> What a good idea. I have thought of it before and posted a feature request at https://bugzilla.mozilla.org/show_bug.cgi?id=1425479
> Unfortunately the feature request was denied. If you post a new bug that shows use cases and why alternatives aren't a good fit, the feature request can be re-evaluated.

Also, the link to your original GitHub issue in there has quite valuable implementation details so I'd like link to it as well: [https://github.com/Rob--W/open-in-browser/issues/5](https://github.com/Rob--W/open-in-browser/issues/5)

I'd like to discuss this issue more. I think - assuming there is enough use of `webRequest` filtering in general - that there is a strong case for a bug because it is still nigh on impossible to do basic text replacement. I know just enough now of the complexities to understand also why it is such a pain for the API - basically the core question is: at what point is the final MIME type known? I think given the complexity of the process - and the fact that generally the browser wholly owns and controls it - having a scenario where one or more addons are trying to interact with it will be tricky. It's not quite the "halting problem" but I'm pretty sure something to do with MIME type detection could be Turing complete.

As I see it, there are many possible depths of solutions that could work to varying degrees:

1. The status quo. Push all detection/re-encoding on to the addon. This seems bad because the logic is complicated and duplicates what the browser is already doing - poorly. Coverage: 0%
2. Provide a solution that does detection solely based on headers - as far as the addon is concerned MIME type would be fixed after that. Also, extend `TextEncoder` so that it can re-encode to the detected MIME type to avoid requiring Content-Type mutations. This would allow for basically one-line implementations, but would be incorrect for situations where the MIME type is in the meta or is sniffed. Coverage: maybe 98%?
3. Delay processing by the addon until the content type is sniffed up to the 512 bytes. Unfortunately I think the addon will likely still want to change e.g. `Content-Type` header based on this, so that sounds like a potential headache for implementation. Otherwise I think something like an `onContentTypeSniffed` as a new hook (or perhaps as an option for `onHeadersReceived`) could be tremendously valuable. Yes, it would also cause problems of a sort if multiple addons chained this. So be it. Coverage: maybe 99.5%?
4. Expose the current MIME type as the browser continues detection as well as the ability to instruct the browser to change MIME type along the way. Complexity? High. Difficulty to implement? Also probably high. Coverage: maybe 99.9%?

So I'd probably lean towards something more in the vein of option 2 or option 3; I think option 3 has the potential to be more problematic from an implementation standpoint.

After pondering the problem for another 3 years, what thoughts have you had on the matter?

wingman-jr-addon:
Also, I did a little research tonight. I searched GitHub for "TextEncoder" and "filterResponseData" and surveyed many of the results. Not exactly the most scientific - and I likely have errors in my hasty analysis - but here are my rough findings:

- 83 source files were examined. I tried to not do too many duplicates, and avoided e.g. forks of FF unit tests.
- 3 of the sources seemed to have recognize and fix both decoding and encoding based on charset etc. (uBlock, Scriptlet Doctor, Adgard)
- 2 other sources had at least partial mitigations for both decoding and encoding. (LibreJS, Header Editor)
- 2 other sources recognized the issue but did not attempt to fix it.
- 18 or so seem are direct copy/modifies the http-response example or inherit the `.disconnect` logic in the first `.ondata` currently present in this example.
- Many sources did not do HTTP status filtering.
- Many sources did decode using stream but did not flush in e.g. `.onstop`

Those are rather abysmal results, but not surprising.
[AnalyzeWebRequestUsage.zip](https://github.com/mdn/webextensions-examples/files/5544331/AnalyzeWebRequestUsage.zip)

This indicates to me that there is a demonstrated need for awareness as well as mechanisms to make some level of correctness easier - character encoding seems like something that should be a minor footnote for the average addon developer.

rebloor:
@Rob--W I know this will be testing the memory a little but, based on your comment, are you happy about the example? If you are, I will test and make sure it works as expected and then merge the changes.

wingman-jr-addon:
@rebloor @Rob--W I saw that activity popped up on this and thought I'd chime in a bit as I've gotten more experience with the issue as my addon has matured. Unfortunately it hasn't translated to the work here (yet?) as I thought this effort was dead. If there's interest I can take another crack at freshening this up.

In short, I now think that probably the best way to handle the issue is to provide a special linked pair of TextDecoder/TextEncoder that hide the detection altogether and allow the developer to just feed in the stream, get a string, and reencode the string. You still need the raw bytes in many cases but additionally for text handling a built-in like this would help greatly. 

The big reason for this approach is primarily because I've now seen many of the different methods that can _in theory_ show up actually _do_ get reported as bugs - including more difficult ones like the `<meta>` approach which requires sniffing the stream. Handling the work to ensure the encoder and decoder stay in sync as the detected type changes is surely something better left to the browser.

 I'd also be curious to hear how @Rob--W 's thoughts have evolved now over the past 6 years since his original feature request.

Rob--W:
> @rebloor @Rob--W I saw that activity popped up on this and thought I'd chime in a bit as I've gotten more experience with the issue as my addon has matured. Unfortunately it hasn't translated to the work here (yet?) as I thought this effort was dead. If there's interest I can take another crack at freshening this up.

Oops, this fell through the cracks. I'll perform a review once this is freshed up.


(The rest of the text below is part of the ongoing discussion about a feature request for Firefox; you can stop reading if you are only interested in the PR)

> In short, I now think that probably the best way to handle the issue is to provide a special linked pair of TextDecoder/TextEncoder that hide the detection altogether and allow the developer to just feed in the stream, get a string, and reencode the string. You still need the raw bytes in many cases but additionally for text handling a built-in like this would help greatly. 

This sounds very convenient from the extension developer's point of view, but a huge burden to the extension framework. Due to content sniffing, the target encoding may sometimes be dependent on the data itself (e.g. binary vs something else). And sometimes there is no lossless way to encode one in another.

> The big reason for this approach is primarily because I've now seen many of the different methods that can _in theory_ show up actually _do_ get reported as bugs - including more difficult ones like the `<meta>` approach which requires sniffing the stream. Handling the work to ensure the encoder and decoder stay in sync as the detected type changes is surely something better left to the browser.

I suppose that you are primarily interested in replacing specific parts of the content without changing the encoding. When the byte sequence is unambiguous (the keyword and replacement), you could also operate on the bytes directly, without using TextEncoder/TextDecoder.


> (One further note/update: not having TextEncoder support encoding into all the encodings TextDecoder can support is a frustrating limitation as I can't simply sniff the charset, decode, and then encode cleanly - this is part of what would be nice with the TextEncoder pair being linked.)

Non-utf8 has little utility in web browsers, and having encoders for non-utf8 encodings would bloat the implementation and result in a larger application binary, as well as a larger attack surface. So while I understand the convenience to developers again, I also recognize the issue with having this functionality natively. Note - if you really want to there is of course nothing stopping you from having encoders the other way around again, but that's probably not optimal.

>  I'd also be curious to hear how @Rob--W 's thoughts have evolved now over the past 6 years since his original feature request.

On https://bugzilla.mozilla.org/show_bug.cgi?id=1425479 ? When I filed that issue I was an external contributor with experience as an extension developer, and browser engineering in Firefox and Chromium (as a volunteer). Now I have been working for a few years as a browser engineer at Mozilla, and one thing that stands out is that there are significantly more ideas than time available to implement them all. Therefore even if an idea on its own seems useful, it may not make the cut due to prioritization. Especially if the feature is niche but requires considerable engineering effort and maintenance.

rebloor:
@wingman-jr-addon please ping Rob when you've had the opportunity to address his feedback and make any other changes.

PR #465, opened 2021-01-19
test for windows,can run,add a picture for the reg key

jinwenmao:


Rob--W:
This PR changes the obvious dummy values to something specific to your local machine, and is not an improvement of the example. Therefore I'm closing this PR.

If you'd like to get a better understanding of the expected format, see https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_manifests

PR #469, opened 2021-05-26
Update documentation of native-messaging example

cwaldren:
Instructs user to view console output via "Inspect" button on extensions page, since there was no output using the previous instructions.

Additionally, updates expected output.


Rob--W (approved):
Thanks for the fix!

PR #478, opened 2021-09-08
Pass -S to env, -u to python

guest271314:
- [23.2.2 -S/--split-string usage in scripts](https://www.gnu.org/software/coreutils/manual/html_node/env-invocation.html#g_t_002dS_002f_002d_002dsplit_002dstring-usage-in-scripts)
- [-u](https://docs.python.org/3/using/cmdline.html#cmdoption-u)


Rob--W:
Could you explain why it was necessary to add this flag? The sample is meant to be a clear example, if the flag is critical for the functionality, some explanatory documentation would be nice.

guest271314:
See links at https://github.com/mdn/webextensions-examples/pull/478#issue-729293651,  https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_messaging

> The batch file then invokes the Python script:
> 
> ```
> @echo off
> 
> python -u "c:\\path\\to\\native-messaging\\app\\ping_pong.py"
> ```

<br>

> This is the Python 2 version:
> 
> ```
> #!/usr/bin/python -u
> 
> # Note that running python with the `-u` flag is required on Windows,
> # in order to ensure that stdin and stdout are opened in binary, rather
> # than text, mode.
> ```

<br>

> In Python 3, the received binary data must be decoded into a string. The content to be sent back to the addon must be encoded into binary data using a struct:
> 
> ```
> #!/usr/bin/python -u
> 
> # Note that running python with the `-u` flag is required on Windows,
> # in order to ensure that stdin and stdout are opened in binary, rather
> # than text, mode.
> ```

PR #483, opened 2022-03-08
Update examples.json

aduntong:
Fix typo


Rob--W:
Thanks for spotting and fixing the typo.

PR #484, opened 2022-03-12
Replace var with let in examples

rebloor:
#### Summary
This change replaces the use of the `var` with `let` in the examples. This addresses the requirement identified in "Replace var with let in code snippets and examples" [#11343](https://github.com/mdn/content/issues/11343) (which in turn was originally identified in discussions on "Use const in place of var" [#11135](https://github.com/mdn/content/pull/11135)).

After completing the global search and replace, a selection of examples was loaded and tested (approximately 10). Only [store-collected-images](https://github.com/mdn/webextensions-examples/tree/master/store-collected-images) was found to have issues, which were corrected by reverting the replace  of`var` with `let` in store-collected-images/webextension-plain/deps/uuidv4.js

### DIFF
--- a/store-collected-images/webextension-plain/deps/idb-file-storage.js
+++ b/store-collected-images/webextension-plain/deps/idb-file-storage.js
@@ -4,7 +4,7 @@
   } else if (typeof exports !== "undefined") {
     factory(exports);
   } else {
-    let mod = {
+    var mod = {
       exports: {}
     };
     factory(mod.exports);
--- a/store-collected-images/webextension-plain/deps/idb-file-storage.js
+++ b/store-collected-images/webextension-plain/deps/idb-file-storage.js
@@ -4,7 +4,7 @@
   } else if (typeof exports !== "undefined") {
     factory(exports);
   } else {
-    var mod = {
+    let mod = {
       exports: {}
     };
     factory(mod.exports);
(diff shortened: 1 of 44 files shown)


store-collected-images/webextension-with-webpack/webpack.config.js
```diff
@@ -41,7 +41,7 @@ module.exports = {
   },
   plugins: [
     // Since some NodeJS modules expect to be running in Node, it is helpful
-    // to set this environment var to avoid reference errors.
+    // to set this environment let to avoid reference errors.
```
rpl, line 44: this should be changed, the inline comment refers to OS environment variables.

store-collected-images/webextension-plain/deps/idb-file-storage.js
```diff
@@ -4,7 +4,7 @@
   } else if (typeof exports !== "undefined") {
     factory(exports);
   } else {
-    var mod = {
+    let mod = {
```
rpl, line 7: I think anything that is third party dependency for a webextensions-example (as in "coming from an outside repo") should be kept out of this change and left unmodified (and as a side note this part is not even part of that library, it is actually code injected automatically as part of the bundling).

rpl (changes requested):
Hi @rebloor, I just took a look (to the diffs, I haven't expanded the context of all of them) and most of them looks ok, only two changes highlighted in the inline review comments below seems something that should be reverted.

rebloor:
@rpl thank you, I have reverted those changes.

PR #491, opened 2022-05-06
Update the `commands` sample to have 2 commands

gurjeet:
Update the `commands` sample to have 2 commands

Before the change, the sample showed how to add one command, and how to
update it, reset it, and respond to it. But usually the extension
developer needs to configure more than one command, and the code was not
readily usable for such a need.

After this patch, the developer can easily see what are the different
places they need to make conditional decisions, to add a new command to
the extension.

This patch also includes some cosmetic changes, like wrapping the long
lines of the README.md, as intended by the authors of Markdown syntax.
If desirable, I can put such cosmetic changes in a separate commit.


rebloor:
@gurjeet I've tested the update, and this extension no longer seems to work as expected, it doesn't pre-populate the existing value for shortcut one, and I am not able to register any new keyboard shortcuts. Could you also explain why you added the description fields? Also, please revert the changes to the readme file, there's no need to manually apply line wraps.

rebloor:
This PR has been inactive for over a year. While the change may be helpful, there are issues in the code and further improvements are needed to achieve the intended goal.

PR #493, opened 2022-05-22
Minor fix in commands extension

sayanarijit:


Rob--W (approved):
Thanks for the improvement!

PR #494, opened 2022-06-11
Changed to runtime for beastify extension when getting content url

interysy:
Hello,
although this doesn't stop extension execution, I found using browser.runtime.getURL() instead of browser.extension.getURL() doesn't trigger an error in the browser console stating " Error: Please use $(ref:runtime.getURL) ". I have changed the code to make use of the recommended code within the beastify example.


Rob--W (changes requested):
The suggestion sounds good. Could you undo the whitespace changes?

interysy:
Should be back to the state before my first commit except the change.

Rob--W (approved):

PR #500, opened 2022-10-10
Updated webpack version in webpack-modules example due to security vulnerabilities

panasenco:
~Also removed deprecated [left-pad](https://www.npmjs.com/package/left-pad) dependency.~ <- Added back in.

### DIFF
--- a/webpack-modules/package.json
+++ b/webpack-modules/package.json
@@ -8,7 +8,8 @@
   },
   "license": "MPL-2.0",
   "devDependencies": {
-    "webpack": "^2.3.2"
+    "webpack": "^5.74.0",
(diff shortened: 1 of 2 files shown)


rebloor:
@panasenco given that the read me describes this as a "minimal example of how to use webpack modules so they can be used in a WebExtension. The example package used by this extension is left-pad" doesn't removing left-pad defeat the point of the example?

panasenco:
That's a good point, @rebloor... I just force-pushed a version that just updates webpack and does nothing else. Tested in Firefox, seems to work.

rebloor:
Thank you @panasenco, tested myself and all looks OK.

PR #501, opened 2022-10-15
Make background script of context-menu-copy-link-with-types non-persistent?

panasenco:
Mozilla documentation [recommends changing background scripts to non-persistent](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Background_scripts#convert_to_non-persistent) to ensure future compatibility with Manifest V3.

I'm pretty sure the example `context-menu-copy-link-with-types` can be made non-persistent as-is. I tested in Firefox and all the functionality seems to work after switching it to non-persistent.

However, I'm not super familiar with background script persistence, so if there are gotchas I missed, let me know!


Rob--W (commented):
The contextMenus.create call in the background page is going to result in an inresolved error, when it wakes up, because the menu will be registered twice. To avoid logspam, check the error by reading the lastError property.

This can be done by adding a callback with `void browser.runtime.lastError;` and adding a comment before that explaining why it's needed.

panasenco:
I just realized that it wasn't really working for me anyway because I'm still on Firefox 105 (from Flathub). Checking the browser console with Ctrl+Shift+J, I get the error "Event pages are not currently supported." Apparently Firefox 106 supports event pages though, so I just need to wait a little bit.

I also tried to get some sort of error from registering the same context menu twice using the below code:
```
try {
    browser.contextMenus.create({
        id: "copy-link-to-clipboard",
        title: "Copy link to clipboard",
        contexts: ["link"],
    });
    console.log("Created context menu");
    browser.contextMenus.create({
        id: "copy-link-to-clipboard",
        title: "Copy link to clipboard",
        contexts: ["link"],
    });
    console.log("Created context menu... again!");
} catch(error) {
    console.error(error);
}
```
However, there's no error at all from the duplicate registration - in either the console or the Parent process Browser Console. I see both log messages just fine.
image

Rob--W:
> I finally found the error message you were referring to in the Multiprocess Browser Console (Ctrl+Shift+J -> Multiprocess): image

Didn't you see this error in the console of the background script? `about:debugging` -> This Firefox -> Inspect button at the extension.

> I also realized that event pages (non-persistent background scripts) are not supported on current stable FF (105), so I added the "strict_min_version" setting and set it 106.0. Haven't tested yet on FF 106 myself yet though...

Even if event pages are not supported, Firefox 105 should still be able to load the extension (and just print a message in the console), right?

The functionality is still the same, an event page is a background page that may be suspended. An event page that never suspends is equivalent to a background page.

panasenco:
> Didn't you see this error in the console of the background script? `about:debugging` -> This Firefox -> Inspect button at the extension.

Nope, didn't see it there at all. It might be related to [bug 1410932](https://bugzilla.mozilla.org/show_bug.cgi?id=1410932).

> Even if event pages are not supported, Firefox 105 should still be able to load the extension (and just print a message in the console), right?
> 
> The functionality is still the same, an event page is a background page that may be suspended. An event page that never suspends is equivalent to a background page.

Yup, it works on 105, no problem.

context-menu-copy-link-with-types/background.js
```diff
@@ -1,8 +1,12 @@
...
 browser.contextMenus.create({
-    id: "copy-link-to-clipboard",
-    title: "Copy link to clipboard",
-    contexts: ["link"],
-});
+        id: "copy-link-to-clipboard",
+        title: "Copy link to clipboard",
+        contexts: ["link"],
+    },
+    () => void browser.runtime.lastError),
```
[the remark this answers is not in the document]
panasenco: Are the comments on lines 1 and 2 insufficient?

context-menu-copy-link-with-types/background.js
```diff
@@ -1,8 +1,12 @@
+// Callback clears runtime.lastError to prevent logspam errors from the menu
```
Rob--W, line 1: Please change it to:
"Callback reads runtime.lastError to prevent an unchecked error from being logged when the extension attempt to register the already-registered menu again. Menu registrations in event pages persist across extension restarts."

120210313069 (approved):

120210313069 (approved):

context-menu-copy-link-with-types/background.js
```diff
@@ -1,8 +1,12 @@
...
+// item being registered twice in case the background script is reloaded.
```
rebloor, line 2:
```suggestion
// Callback reads runtime.lastError to prevent an unchecked error from being 
// logged when the extension attempt to register the already-registered menu 
//again. Menu registrations in event pages persist across extension restarts.
```

context-menu-copy-link-with-types/manifest.json
```diff
@@ -3,12 +3,18 @@
     "name": "Context menu: Copy link with types",
     "description": "Add a context menu option to links to copy the link to the clipboard, as plain text and as a link in rich HTML.",
     "version": "1.0",
+    "browser_specific_settings": {
+        "gecko": {
+            "strict_min_version": "106.0"
+        }
+    },
```
rebloor, line 10:
```suggestion
```

context-menu-copy-link-with-types/background.js
```diff
@@ -1,8 +1,16 @@
+// Callback reads runtime.lastError to prevent an unchecked error from being 
+// logged when the extension attempt to register the already-registered menu 
+//again. Menu registrations in event pages persist across extension restarts.
```
Rob--W, line 3:
```suggestion
// again. Menu registrations in event pages persist across extension restarts.
```

Rob--W (approved):

rebloor:
@panasenco many thanks for kicking this off

PR #502, opened 2022-10-23
Don't use <div /> as a button

Eik-S:
As suggested by @rebloor this alignes the code to my changes in [the tutorial to this example](https://github.com/mdn/content/pull/21410)

PR #505, opened 2022-12-02
Fix beastify code

aloopuri:
#502 introduced an error where if you click on the background of the popup or if you added anything else to the popup page and clicked on it would try and run the beastify function, and this would cause the screen to go blank.

This issue was this else statement which accepted any and all clicks on the popup page.

https://github.com/mdn/webextensions-examples/blob/6d8f07243a23c6b402cc33c7ac926155f524f3ce/beastify/popup/choose_beast.js#L68-L78

### DIFF
--- a/beastify/popup/choose_beast.css
+++ b/beastify/popup/choose_beast.css
@@ -16,7 +16,7 @@ html, body {
   cursor: pointer;
 }
 
-.button:hover {
+.beast:hover {
   background-color: #CFF2F2;
 }
 
--- a/beastify/popup/choose_beast.css
+++ b/beastify/popup/choose_beast.css
@@ -6,7 +6,7 @@ html, body {
   display: none;
 }
 
-button {
+.button {
   border: none;
   width: 100%;
   margin: 3% auto;
@@ -14,17 +14,20 @@ button {
   text-align: center;
   font-size: 1.5em;
   cursor: pointer;
-  background-color: #E5F2F2;
 }
 
-button:hover {
+.button:hover {
   background-color: #CFF2F2;
 }
 
-button[type="reset"] {
+.beast {
+  background-color: #E5F2F2;
+}
+
+.reset {
   background-color: #FBFBC9;
 }
 
-button[type="reset"]:hover {
+.reset:hover {
   background-color: #EAEA9D;
 }
--- a/beastify/popup/choose_beast.html
+++ b/beastify/popup/choose_beast.html
@@ -8,10 +8,10 @@
 
 <body>
   <div id="popup-content">
-    <button>Frog</button>
-    <button>Turtle</button>
-    <button>Snake</button>
-    <button type="reset">Reset</button>
+    <button class="button beast">Frog</button>
+    <button class="button beast">Turtle</button>
+    <button class="button beast">Snake</button>
+    <button class="button reset">Reset</button>
   </div>
   <div id="error-content" class="hidden">
     <p>Can't beastify this web page.</p><p>Try a different page.</p>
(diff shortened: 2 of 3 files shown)


beastify/popup/choose_beast.html
```diff
@@ -8,10 +8,10 @@
 
 <body>
   <div id="popup-content">
-    <button>Frog</button>
-    <button>Turtle</button>
-    <button>Snake</button>
-    <button type="reset">Reset</button>
+    <button class="button beast">Frog</button>
```
estelle, line 11: this shouldn't be necessary. Just style based on `<button>` and `[type="reset"]`

beastify/popup/choose_beast.css
```diff
@@ -6,25 +6,28 @@ html, body {
...
-  background-color: #E5F2F2;
 }
 
-button:hover {
+.beast:hover {
   background-color: #CFF2F2;
 }
 
-button[type="reset"] {
+.beast {
```
estelle, line 23: is there a reason we to add classes? Is there something about extensions that make type selectors and attribute selectors not work??

estelle (commented):
This PR adds several class names that don't seem necessary. Are they?

beastify/popup/choose_beast.css
estelle: Or is this just reverting 502?

aloopuri:
I added a new fix. I reverted back to using types again as you suggested and added a type to the other buttons too, which fixes the issue I originally mentioned (i.e. the else statement allowing all clicks and showing a blank page when the background is clicked). Also it's good practice as mentioned [here](https://dev.to/clairecodes/why-its-important-to-give-your-html-button-a-type-58k9)

I thought of using classes again because of the event handler for the entire page. If someone was to extend this project and add more buttons with different functionality (such as transforming the images), we would need another way of identifying what those buttons are, as identifying them based on only their type attribute wouldn't be possible, would it? But I guess that may be beyond the scope of this project anyway.

beastify/popup/choose_beast.html
```diff
@@ -8,9 +8,9 @@
 
 <body>
   <div id="popup-content">
-    <button>Frog</button>
-    <button>Turtle</button>
-    <button>Snake</button>
+    <button type="button">Frog</button>
```
Rob--W, line 11: `type="button"` is not explicitly needed. By default it's `type="submit"`.

beastify/popup/choose_beast.js
```diff
@@ -65,13 +65,13 @@ function listenForClicks() {
      * Get the active tab,
      * then call "beastify()" or "reset()" as appropriate.
      */
-    if (e.target.type === "reset") {
+    if (e.target.type === "button") {
```
Rob--W, line 68: Instead of being indirect here, let's just prepend:

```lang=javascript
if (e.target.tagName !== "BUTTON" || !e.target.closest("#popup-content")) {
  // Ignore when click is not on a button within <div id="popup-content">.
  return;
}
```

beastify/popup/choose_beast.js
```diff
@@ -65,7 +65,10 @@ function listenForClicks() {
      * Get the active tab,
      * then call "beastify()" or "reset()" as appropriate.
      */
-    if (e.target.type === "reset") {
+    if (e.target.tagName !== "BUTTON" || !e.target.closest("#popup-content")) {
+      // Ignore when click is not on a button within <div id="popup-content">.
+      return;
+    } else if (e.target.type === "reset") {
```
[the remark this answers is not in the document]
aloopuri: Oh ok I'll fix it now

Rob--W (approved):
Thanks for the fix!

Issue #509, opened 2023-01-01
Python Native Messaging host does not process 1MB input


guest271314:
The  example Python Native Messaging host should be able to process input up to 1MB. The host exits when input is 

```
new Array(174800)
```

where
```
JSON.stringify(new Array(174800)).length // 874001
``` 

The Native Messaging host should be able to process at least up and including 

```
port.postMessage(new Array(209000))
``` 

where

```
JSON.stringify(new Array(209000)).length // 1045001
```

is still less than 

```
1024*1024 // 1048576
```

Rob--W:
What makes you believe that only 1 MB is accepted? When I load the example on Linux, I can successfully continue to post to the native messaging app even after sending 1 MB of data.

```js
// port was defined in background.js via connectNative + onMessage listener.

// Send 5 MB data. The message is ignored because the example only responds to "ping".
port.postMessage(new Array(1024 * 1024 * 5));

port.postMessage("ping");
// Result: "pong3" (after waiting a short while for the previous message to have been flushed)
```

guest271314:
> The addition of separators is not meaningful at all

Yes, it is. The omission means when we do

```
port.postMessage(port.postMessage(new Array(174763));
```

the current Python Native Messaging host exits and this

```
def encodeMessage(messageContent):
        encodedContent = json.dumps(messageContent).encode('utf-8')
        f = open('data.txt', 'w')
        f.write(str(len(encodedContent)))
        f.close()
        encodedLength = struct.pack('@I', len(encodedContent))
        return {'length': encodedLength, 'content': encodedContent}
```
will write 

```
1048578
```

to data.txt, greater than the Native Messaging protocol 1024*1024 - when the length _should_ be 

```
873816
```

```
JSON.stringify(new Array(174763)).length // 873816
```

which means we cannot pass and get echoed the maximum Native Messaging protocol message from client to host, 1024*1024 with `separators()` omitted.

This

```
port.postMessage(new Array(209715));
``` 

will always exit in the Python script you have in your current code, because the default JSON formatting space character that Python adds, see https://docs.python.org/3/library/json.html

> If specified, _separators_ should be an (item_separator, key_separator) tuple. The default is `(', ', ': ')` if indent is `None` and `(',', ': ')` otherwise. To get the most compact JSON representation, you should specify `(',', ':')` to eliminate whitespace.

[...]

PR #510, opened 2023-01-07
Eliminate space counted as message length

guest271314:
When separators=(',', ':') is not passed to json.dumps() parameter the length for input from JavaScript (which is serialized to a JSON-like format by the application (Chrome))port.postMessage(new Array(174763)); is read in Python as

1048578

which is greater than 1024*1024. Essentially space character formatting were being encoded as part of the length of the JSON.

Solved by passing separators=(',', ':') for “compact encoding” json — JSON encoder and decoder — Python 3.11.1 documentation.

Now we can write 1MB with

port.postMessage(new Array(209715));

from JavaScript and omit including and counting formatting space characters as part of encoded message length.

### DIFF
--- a/native-messaging/app/ping_pong.py
+++ b/native-messaging/app/ping_pong.py
@@ -18,6 +18,7 @@ try:
     # Encode a message for transmission,
     # given its content.
     def encodeMessage(messageContent):
+        # https://github.com/mdn/webextensions-examples/issues/509
         # https://discuss.python.org/t/how-to-read-1mb-of-input-from-stdin/22534/19
         # https://stackoverflow.com/a/56563264
         # https://docs.python.org/3/library/json.html#basic-usage
@@ -51,6 +52,7 @@ except AttributeError:
     # Encode a message for transmission,
     # given its content.
     def encodeMessage(messageContent):
+        # https://github.com/mdn/webextensions-examples/issues/509
         # https://discuss.python.org/t/how-to-read-1mb-of-input-from-stdin/22534/19
         # https://stackoverflow.com/a/56563264
         # https://docs.python.org/3/library/json.html#basic-usage
--- a/native-messaging/app/ping_pong.py
+++ b/native-messaging/app/ping_pong.py
@@ -21,7 +21,8 @@ try:
         # https://docs.python.org/3/library/json.html#basic-usage
         # To get the most compact JSON representation, you should specify 
         # (',', ':') to eliminate whitespace.
-        # We want the most compact representation because the browser rejects # messages that exceed 1 MB.
+        # We want the most compact representation because the browser rejects
+        # messages that exceed 1 MB.
         encodedContent = json.dumps(messageContent, separators=(',', ':')).encode('utf-8')
         encodedLength = struct.pack('@I', len(encodedContent))
         return {'length': encodedLength, 'content': encodedContent}
--- a/native-messaging/app/ping_pong.py
+++ b/native-messaging/app/ping_pong.py
@@ -18,6 +18,11 @@ try:
     # Encode a message for transmission,
     # given its content.
     def encodeMessage(messageContent):
+        # https://discuss.python.org/t/how-to-read-1mb-of-input-from-stdin/22534/19
+        # https://stackoverflow.com/a/56563264
+        # https://docs.python.org/3/library/json.html#basic-usage
+        # To get the most compact JSON representation, you should specify 
+        # (',', ':') to eliminate whitespace.
         encodedContent = json.dumps(messageContent, separators=(',', ':')).encode('utf-8')
         encodedLength = struct.pack('@I', len(encodedContent))
         return {'length': encodedLength, 'content': encodedContent}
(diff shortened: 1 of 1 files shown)


guest271314:
Fixes https://github.com/mdn/webextensions-examples/issues/509

Rob--W:
The ping_pong example only responds to "pong". The addition of `separators` is not meaningful at all, since the protocol doesn't care whether there is a space or not. I'll therefore close this PR, and follow up on the original issue to figure out what you're doing.

guest271314:
> The ping_pong example only responds to "ping" with "pong". The addition of `separators` is not meaningful at all, since the protocol doesn't care whether there is a space or not. I'll therefore close this PR, and follow up on the original issue to figure out what you're doing.

Yes, it is.

Your current Python Native Messaging host script has a bug. In ints current form it cannot process the maxium capable input per Native Messaging protocol, 1024*1024. The space character added by Python `json.dumps()` by deafault will make your current script exit when input is 

```
port.postMessage(new Array(174763));
```

This by default will add space characters

```
def encodeMessage(messageContent):
        encodedContent = json.dumps(messageContent).encode('utf-8')
        f = open('data.txt', 'w')
        f.write(str(len(encodedContent)))
        f.close()
        encodedLength = struct.pack('@I', len(encodedContent))
        return {'length': encodedLength, 'content': encodedContent}
```

so the length will be

```
1048578
```

when the length _should_ be

```
873816
```

per

```
JSON.stringify(new Array(174763)).length // 873816
```

That space character in the Python script is useless because we get a plain JavaScript object back in JavaScript - not formatted JSON.

https://docs.python.org/3/library/json.html

> If specified, _separators_ should be an (item_separator, key_separator) tuple. The default is `(', ', ': ')` if indent is `None` and `(',', ': ')` otherwise. To get the most compact JSON representation, you should specify `(',', ':')` to eliminate whitespace.

I can pass

```
port.postMessage(new Array(209715));
```

to C, C++, QuickJS, Deno, Node.js Native Messaging hosts 


- https://github.com/guest271314/native-messaging-c
- https://github.com/guest271314/native-messaging-cpp
- https://github.com/guest271314/native-messaging-quickjs
- https://github.com/guest271314/native-messaging-deno
- https://github.com/guest271314/native-messaging-nodejs

and get the complete 1024*1024 message back. Your current Python Native Messaging host has a bug that will exit when



```
port.postMessage(new Array(174763));
```

is executed in JavaScript from the client.

So your current Python Native Messaging host script is broken, though is easily fixed by passing `separators()` to `json.dumps()`.

native-messaging/app/ping_pong.py
```diff
@@ -18,7 +18,7 @@ def getMessage():
     # Encode a message for transmission,
     # given its content.
     def encodeMessage(messageContent):
-        encodedContent = json.dumps(messageContent).encode('utf-8')
+        encodedContent = json.dumps(messageContent, separators=(',', ':')).encode('utf-8')
```
Rob--W, line 21: Please prepend a brief comment to explain that the separators argument exists to have a compact message, to reduce the odds of hitting the 1MB message size limit. Feel free to link to the github issue for context.

Rob--W (changes requested):

guest271314:
Related https://github.com/mdn/content/pull/23451

native-messaging/app/ping_pong.py
```diff
@@ -18,7 +18,13 @@ def getMessage():
     # Encode a message for transmission,
     # given its content.
     def encodeMessage(messageContent):
-        encodedContent = json.dumps(messageContent).encode('utf-8')
+        # https://github.com/mdn/webextensions-examples/issues/509
+        # https://discuss.python.org/t/how-to-read-1mb-of-input-from-stdin/22534/19
+        # https://stackoverflow.com/a/56563264
```
Rob--W, line 23: Let's remove these three lines. The link to the Python documentation below is sufficient. Your GH issue and Python code would be causing more confusion than that it would help, because these all frame the issue as "issue with reading 1MB", while the actual issue is with writing 1 MB from the Python script (or rather, the browser not accepting 1 MB).

native-messaging/app/ping_pong.py
```diff
@@ -18,7 +18,13 @@ def getMessage():
...
+        # https://docs.python.org/3/library/json.html#basic-usage
+        # To get the most compact JSON representation, you should specify 
+        # (',', ':') to eliminate whitespace.
```
Rob--W, line 26: Add this line:

```python
        # We want the most compact representation because the browser rejects
        # messages that exceed 1 MB.
```

Rob--W (changes requested):

native-messaging/app/ping_pong.py
```diff
@@ -18,7 +18,11 @@ def getMessage():
     # Encode a message for transmission,
     # given its content.
     def encodeMessage(messageContent):
-        encodedContent = json.dumps(messageContent).encode('utf-8')
+        # https://docs.python.org/3/library/json.html#basic-usage
+        # To get the most compact JSON representation, you should specify 
+        # (',', ':') to eliminate whitespace.
+        # We want the most compact representation because the browser rejects # messages that exceed 1 MB.
```
Rob--W, line 24:
```suggestion
        # We want the most compact representation because the browser rejects
        # messages that exceed 1 MB.
```

native-messaging/app/ping_pong.py
```diff
@@ -46,7 +50,11 @@ def getMessage():
     # Encode a message for transmission,
     # given its content.
     def encodeMessage(messageContent):
-        encodedContent = json.dumps(messageContent)
+        # https://docs.python.org/3/library/json.html#basic-usage
+        # To get the most compact JSON representation, you should specify 
+        # (',', ':') to eliminate whitespace.
+        # We want the most compact representation because the browser rejects # messages that exceed 1 MB.
```
Rob--W, line 56:
```suggestion
        # We want the most compact representation because the browser rejects
        # messages that exceed 1 MB.
```

Rob--W (commented):
Almost ready to merge

native-messaging/app/ping_pong.py
Rob--W: This one still needs to be addressed.

Rob--W (approved):
Thanks!

PR #523, opened 2023-04-17
Code now reflects newer versions of browser.menus API

Pedro-Celeste:
### Description

The background script of `menu-demo` was updated to use a newer version of the `browser.menus` API, so the code doesn't need to create an extra variable `checkedState` to keep track of menu item state.

Code contains more details.

### Motivation

The previous version of the code contained a  comment saying that a feature is not supported by the Firefox version of the `browser.menus` API.

Since now it is supported, I think it should be implemented.

This way, new developers learning the API won't be confused by the information mismatch in the [documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus/OnClickData) and the example extension.

Awesome extension by the way, it helped me a lot to understand this API.

### DIFF
--- a/menu-demo/background.js
+++ b/menu-demo/background.js
@@ -77,14 +77,12 @@ browser.menus.create({
   contexts: ["all"]
 }, onCreated);
 
-let checkedState = true;
-
 browser.menus.create({
   id: "check-uncheck",
   type: "checkbox",
   title: browser.i18n.getMessage("menuItemUncheckMe"),
   contexts: ["all"],
-  checked: checkedState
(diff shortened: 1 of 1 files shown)


rebloor (approved):
Loaded revised extension. Ran and noted it behaved as documented. Check code changes, all looked OK.

rebloor:
Many thanks for spotting and fixing that @Pedro-Celeste

Pedro-Celeste:
@rebloor No problem! :)

PR #537, opened 2023-08-22
window-manipulator update for issue 431

### LINKED ISSUE #431: windows.create() promise will never be fulfilled
Snippet from this example:
https://github.com/mdn/webextensions-examples/blob/master/window-manipulator/window.js#L28
```js
let creating = browser.windows.create(createData);
  creating.then(() => {
  console.log("The normal window has been created");
});
```
I cant see `console.log`  message while i am inspecting this extension. 

I only see such messages:
```
Navigated to chrome://devtools/content/shared/webextension-fallback.html
Webconsole context has changed
```

It seems that this promise will never be fullfiled. Looks like a bug in WebExtensions API because [examples from MDN ](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/windows/create) don't work too.

rebloor:
### Description

Amends the extension to use a sidebar so that the console output indicating a normal window has been created is guaranteed to work.

Fixes #431


Rumyra (approved):

zombie (approved):
Thanks!

PR #541, opened 2023-10-01
Reverse var to let change

### LINKED ISSUE #536: Broken root-cert-stats due to var -> let change
### What information was incorrect, unhelpful, or incomplete?

The current root-cert-stats extension throws a "TypeError: can't convert undefined to object" at popup.js 8:22
The popup window always shows "No data to display yet"

About a year ago, someone changed var to let at line 3 of background.js for the variable rootCertStats.  As a result, backgroundWindow.rootCertStats is undefined.

Changing the let back to var makes things work again.

### What did you expect to see?

No error should be thrown.  The popup window should show the stats.

### Do you have any supporting links, references, or citations?

_No response_

### Do you have anything more you want to share?

It appears the var -> let change was done without testing.  I wonder how many other extensions were broken.

rebloor:
### Description

Fixes #536 (reverses the change made in https://github.com/mdn/webextensions-examples/pull/484)


root-cert-stats/background.js
```diff
@@ -1,6 +1,6 @@
 "use strict";
 
-let rootCertStats = {};
+var rootCertStats = {};
```
Rob--W, line 3:
```suggestion
// Note: declared with "var" because popup.js references this global variable.
// If this were to be declared with "const" or "let", then the variable would
// still be available to this file, but not to popup.js.
var rootCertStats = {};
```

Rob--W (approved):
Please include the comment before merging.

PR #554, opened 2024-02-09
Fix the example with Firefox on macOS when managed storage isn't available.

Dalzhim:
When managed storage isn't available (i.e. : using Firefox 122 on macOS), the example code throws an exception while trying to restore the value from managed storage and never proceeds to restore the value saved in storage.sync.

This change handles the exception so that the example works out of the box.

### DIFF
--- a/favourite-colour/options.js
+++ b/favourite-colour/options.js
@@ -6,8 +6,12 @@ async function saveOptions(e) {
 }
 
 async function restoreOptions() {
-  let res = await browser.storage.managed.get('colour');
-  document.querySelector("#managed-colour").innerText = res.colour;
+	try {
+		let res = await browser.storage.managed.get('colour');
+		document.querySelector("#managed-colour").innerText = res.colour;
+	} catch(error) {
+		console.log(JSON.stringify(error));
+	}
 
   res = await browser.storage.sync.get('colour');
   document.querySelector("#colour").value = res.colour || 'Firefox red';


rebloor:
@Dalzhim could you provide some more information on why you think this changing necessary, managed storage has been available in Firefox from version 57

Dalzhim:
@rebloor: i used the sample code and it threw the exception. I do not remember the exact circumstances now, but i was probably attempting to use that sample code from within a web extension’s background script or injected script. Nevertheless, managed storage apparently wasn’t available in the context (firefox 122 running on macos according to my original report) and if the error had already been logged, it’d have saved me some time debugging the issue.

rebloor:
@Dalzhim Without more precise details, it's difficult to comment further. If you encounter the issue again or are able to reproduce it otherwise, please raise it on [the forum](https://discourse.mozilla.org/c/add-ons) to help determine whether it's a bug or a usage issue. In the meantime, I'm closing this PR as there don't appear to be any similar known issues with this feature. Thanks.

Dalzhim:
I’m telling you the sample code failed throwing an exception. I believe it was in the context of a web extension. Now if you’d rather keep the sample code the way it is, meaning without proper handling, then suit yourself. I don’t plan on commenting any further.

Rob--W:
@rebloor The PR author here raised a valid point: in Firefox, storage.managed.get throws if the storage manifest has not been specified.

In contrast, in Chrome the default values as specified in the `storage.managed_schema` key is used instead.

This difference is currently not documented on MDN, but a report exists at https://bugzilla.mozilla.org/show_bug.cgi?id=1868153

It would be useful to catch errors and print them somewhere, but with a comment explaining why.

Relevant documentation can be found at:
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/storage
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/managed
- https://bugzilla.mozilla.org/show_bug.cgi?id=1868153

rebloor:
Sorry @Rob--W and @Dalzhim i should have realised that underlying issue would be the extension storage file not, as reported, managed storage being absent on Forefox. @Rob--W can we differentiate between the managed storage file missing compared to the data item?

favourite-colour/options.js
```diff
@@ -6,8 +6,12 @@ async function saveOptions(e) {
...
 
 async function restoreOptions() {
-  let res = await browser.storage.managed.get('colour');
-  document.querySelector("#managed-colour").innerText = res.colour;
+	try {
+		let res = await browser.storage.managed.get('colour');
+		document.querySelector("#managed-colour").innerText = res.colour;
+	} catch(error) {
+		console.log(JSON.stringify(error));
+	}
```
rebloor, line 14: @Rob--W something more like this?
```suggestion
  try {
    let res = await browser.storage.managed.get('colour');
    if (res.colour === undefined) {
      console.error("Data key colour not defined in the managed storage file.");
    }
    document.querySelector("#managed-colour").innerText = res.colour || "Data key colour not defined in the managed storage file.";
  } catch(error) {
    console.error("Managed storage file not found, check it's stored in an appropriate location.");
  }
```

dotproto: @rebloor, I took your suggestion and iterated on it a little.

 ```suggestion
  try {
    let res = await browser.storage.managed.get('colour');
    document.querySelector("#managed-colour").innerText = res.colour || "Could not find 'colour' in managed storage.";
  } catch(error) {
    if (error.message === "Managed storage manifest not found") {
      document.querySelector("#managed-colour").innerText = "Failed to find a managed storage manifest.";
    } else {
      document.querySelector("#managed-colour").innerText = `Unexpected managed storage error: ${error.message}`;
    }
  }
```

**Review by rebloor — APPROVED** [review]

**Comment by rebloor** [comment]
@dotproto any thoughts on why the lint check is failing?

File: favourite-colour/options.js [hunk 2]
```diff
@@ -6,8 +6,16 @@ async function saveOptions(e) {
...
+ try {
+   let res = await browser.storage.managed.get('colour');
+   document.querySelector("#managed-colour").innerText = res.colour || "Could not find 'colour' in managed storage.";
+ } catch(error) {
+   if (error.message === "Managed storage manifest not found") {
+     document.querySelector("#managed-colour").innerText = "Managed storage manifest not found. Make sure it's stored in an appropriate location.";
+   } else {
+     document.querySelector("#managed-colour").innerText = `Unexpected managed storage error: ${error.message}`;
+   }
+ }
```
dotproto: Normalize indentation.

```suggestion
  try {
    let res = await browser.storage.managed.get('colour');
    document.querySelector("#managed-colour").innerText = res.colour || "Could not find 'colour' in managed storage.";
  } catch(error) {
    if (error.message === "Managed storage manifest not found") {
      document.querySelector("#managed-colour").innerText = "Managed storage manifest not found. Make sure it's stored in an appropriate location.";
    } else {
      document.querySelector("#managed-colour").innerText = `Unexpected managed storage error: ${error.message}`;
    }
  }
```

PR #555, opened 2024-03-08
Changed for manifest version 3 and compatibility to Chrome and CCleaner browsers

gum1900:
### Description
I changed the "beastify" extension to 
   1. to run with manifest version 3
   2. to make compatible with Chrome and CC cleaner browsers

### Motivation
The changes demonstrate Mozilla extensions running with manifest version 3.

### Additional details
Unfortunately the change run very well with Chrome and CCleaner browsers, but does not run well with Firefox browser.  On Firefox browser, content JS does not load.  Pop-up and website page Debuggers does not provide any further information except a console message of "undefined: undefined" on the pop-up debugger.  I have raised a bug "https://bugzilla.mozilla.org/show_bug.cgi?id=1884298" (Bug ID: 1884298) with Mozilla.

Nevertheless, the changes can still be helpful for developers writing compatible extensions  with manifest version 3 

Bugzilla bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1884298

### DIFF
--- a/beastify/content_scripts/beastify.js
+++ b/beastify/content_scripts/beastify.js
@@ -1,3 +1,10 @@
+// Making compatible with Chrome
+if (typeof browser == "undefined") {
+  // `browser` is not defined in Chrome, but Manifest V3 extensions in Chrome
+  // also support promises in the `chrome` namespace, like Firefox. To easily
+  // test the example without modifications, polyfill "browser" to "chrome".
+  globalThis.browser = chrome;
+}
 (function() {
   /**
    * Check and set a global guard variable.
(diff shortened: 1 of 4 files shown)


Rob--W:
A mix of MV2 and MV3 is confusing, we should improve the documentation on MDN.

This PR converts an example to MV3, but does not include browser_specific_settings.gecko.id. While Firefox does not require that, the extension store at addons.mozilla.org does require it. I haven't decided whether we should include it in all examples (at the risk of the user trying to submit it to AMO in unmodified form) or to exclude it (at the risk of the user getting stuck at the last minute, when trying to upload to AMO).

Another consideration is that Chrome currently warns when browser_specific_settings is present. This will be fixed in Chrome 131, to not warn any more.

alexmiddeleer:
@Rob--W I am happy to help with this project if desired. Either the documentation side or converting some examples.

PR #558, opened 2024-04-20
native-messaging: fix Windows checker script not to fail on missing HKCU when HKLM is set

### LINKED ISSUE #557: minor bug with native messaging example
this line will throw when you create required key in `HKEY_LOCAL_MACHINE` instead of `HKEY_CURRENT_USER` ( which is allowed ) , so this script should not behave like this :
https://github.com/mdn/webextensions-examples/blob/faadfca8ddce0c02cc20c3c261c76a9b50073122/native-messaging/check_config_win.py#L22

also , as the script is in Python anyway , the whole thing can be simplified using for-else :

```
for root in key_roots:
    try:
        key = winreg.OpenKey(getattr(winreg, root), key_path)
        print('Checking:', root, key_path)
        res = winreg.QueryValueEx(key, '')
    except FileNotFoundError:
        continue
    break
else:
    raise ValueError('Could not find a registry entry, aborting.')
```

let me know if you want me rather create a small PR with this fix

dahn-zk:
### Description

from https://github.com/mdn/webextensions-examples/issues/557 :

> this line will throw when you create required key in `HKEY_LOCAL_MACHINE` instead of `HKEY_CURRENT_USER` ( which is allowed ) , so this script should not behave like this :
> 
> https://github.com/mdn/webextensions-examples/blob/faadfca8ddce0c02cc20c3c261c76a9b50073122/native-messaging/check_config_win.py#L22
> 
> also , as the script is in Python anyway , the whole thing can be simplified using for-else :
> 
> ```
> for root in key_roots:
>     try:
>         key = winreg.OpenKey(getattr(winreg, root), key_path)
>         print('Checking:', root, key_path)
>         res = winreg.QueryValueEx(key, '')
>     except FileNotFoundError:
>         continue
>     break
> else:
>     raise ValueError('Could not find a registry entry, aborting.')
> ```

### Motivation

i was bored

### Additional details

update : ignore the following , i created a PR for that https://github.com/mdn/content/pull/33211

btw , in the docs https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_messaging#registry there is an incorrect info about registry keys . it says "two registry entries should be created for the messaging to work" as if "both" in `HKEY_CURRENT_USER` and `HKEY_LOCAL_MACHINE` . but this is not true . only one should be . here is the proof in the Firefox source code : https://searchfox.org/mozilla-central/source/toolkit/components/extensions/NativeManifests.sys.mjs#59-83

fixes #557

### DIFF
--- a/native-messaging/README.md
+++ b/native-messaging/README.md
@@ -16,7 +16,7 @@ To get this working, there's a little setup to do.
 
 ### Windows setup ###
 
-1. Check you have Python installed, and that your system's PATH environment variable includes the path to Python.  See [Using Python on Windows](https://docs.python.org/2/using/windows.html). You'll need to restart the web browser after making this change, or the browser won't pick up the new environment variable.
+1. Check you have Python installed, and that your system's PATH environment variable includes the path to Python.  See [Using Python on Windows](https://docs.python.org/3/using/windows.html). You'll need to restart the web browser after making this change, or the browser won't pick up the new environment variable.
 2. Edit the "path" property of "ping_pong.json" to point to the location of "ping_pong_win.bat" on your computer. Note that you'll need to escape the Windows directory separator, like this: `"path": "C:\\Users\\MDN\\native-messaging\\app\\ping_pong_win.bat"`.
 3. Edit "ping_pong_win.bat" to refer to the location of "ping_pong.py" on your computer.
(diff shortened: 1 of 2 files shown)


dotproto (approved):
Thanks for this PR and https://github.com/mdn/content/pull/33211!

PR #559, opened 2024-04-25
migrate `list-cookies` extension to manifest version 3

2-X:
### Description

`browser` --> `chrome`
idk if this is an issue others run into but i assumed it was because of manifest v3

### DIFF
--- a/list-cookies/cookies.js
+++ b/list-cookies/cookies.js
@@ -1,39 +1,40 @@
+function getActiveTab() {
+  chrome.tabs.query({currentWindow: true, active: true}, function(tabs) {
+    showCookiesForTab(tabs);
+  });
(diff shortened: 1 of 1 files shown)


rebloor:
@2-X you haven't made any changes that migrate this to Manifest v3. That would start with a change to the manifest.json file line `"manifest_version": 2`. These examples are intended for firefox, hence they use the `browser` namespace. Were you running this example in Chrome?

2-X:
@rebloor thanks for the reply! yes i was running it in chrome, i see that `browser` is the firefox version of `chrome` API. something like this PR should instead be part of a larger refactor to make them all work on chrome, if desired, so i closed this PR 👍🏼

PR #561, opened 2024-05-11
Illustrative function name

s-c-p:
updateActiveTab has ambiguous meaning, most prominent of which doesn't mean what we are actually trying to do.

### Description

updateActiveTab has ambiguous meaning, most prominent interpretation of which is not what we are actually doing in the function body.

### Motivation

I know JS as server-side engineer (we have very pedantic rules on where we are allowed to change state), in this succinct but cool sample code, the flow-of-control is slightly confusing (if we don't know browser event loop) and I guess **descriptive function names would reduced the confusion for new comers**, the author had context and experience, but for newbies updateActiveTab would mean perhaps we are modifying something on page. Reading function def would clear the misconception, but add mental burden.

### DIFF
--- a/bookmark-it/background.js
+++ b/bookmark-it/background.js
@@ -39,7 +39,7 @@ browser.browserAction.onClicked.addListener(toggleBookmark);
 /*
  * Switches currentTab and currentBookmark to reflect the currently active tab
  */
-function updateActiveTab(tabs) {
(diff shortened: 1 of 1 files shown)


dotproto (approved):
Seems like a good change. Thanks, @s-c-p!

PR #565, opened 2024-07-19
Update for MV3 and Chrome compatibility

dotproto:
Updates the stored-credentials example to use Manifest V3 and to add support for Chrome. This demo now works in Firefox and Chrome. It does not work in Safari because at the moment `webRequestAuthProvider` is unsupported.

### DIFF
--- a/stored-credentials/auth.js
+++ b/stored-credentials/auth.js
@@ -1,3 +1,5 @@
+// Polyfill the "browser" global in Chrome.
+globalThis.browser ??= chrome;
 
 let target = "https://httpbin.org/basic-auth/*";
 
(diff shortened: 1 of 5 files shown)


rebloor:
@dotproto This change makes the example inconsistent with the guidance in the readme file: "These examples are tested in Firefox. They may work in other browsers, if the [browser supports the APIs used](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Browser_support_for_JavaScript_APIs). Note that these examples all use the browser namespace and promises to work with asynchronous functions. This means the examples won't work in Chrome unless you use the [polyfill provided by Mozilla](https://github.com/mozilla/webextension-polyfill). See [the overview of WebExtension APIs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API) for more information." 
I'm wondering whether we should continue with this as an MV2, Firefox-only example unless we have plans to update all the extensions to MV3 and Chrome support. We introduced an inconsistency and potentially an expectation of updates we may not have plans to make.

dotproto:
That's a very good point. I've revised the demo to revert back to MV2 and to remove Chrome compatibility changes. While I *do* want to put concerted effort into updating the demos, we should stay consistent until we establish a plan and update the docs to reflect it.

rebloor (approved):
@dotproto sorry for the delay. I've left one small suggestion. I've also run the extension and confirmed that it behaves as expected.

PR #566, opened 2024-08-12
fix(examples.json): remove &#x200B;

fiji-flo:
### Description

There are two &#x200B; (zero width space) in the
examples.json

Let's remove them


argl (approved):
wow

PR #573, opened 2024-11-18
Allow contributors to lint individual examples

dotproto:
### Description

This PR introduces a new npm script: `lint:cwd`. This script allows contributors to run the linter against their current directory rather than against the entire repo.

### Motivation

 It was frustrating to try to filter `npm run lint` logs for only the files that were relevant to my PR.

### Additional details

N/A

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "scripts": {
     "test": "eslint .",
     "lint": "eslint .",
+    "lint:cwd": "eslint $INIT_CWD",
     "lint:fix": "eslint . --fix"
   },
   "license": "MIT",


Rob--W:
Note that there is an existing `lint:fix` task. I considered asking to duplicate it, but then recalled that you can do something like `npm run lint:cwd -- --fix` to achieve something similar.

Taking it one step further... Do you need this PR at all? Wouldn't `npx eslint .` do what you want (and is also shorter)?

dotproto:
Yes, `npx eslint .` would perform the same operation as the proposed `lint:cwd` script. I tend to think of npm scripts as the canonical commands a developer would use when contributing to a project; after you clone a new project, you can execute `npm run` to see what kind of scripts are provided. One useful function scripts provide in this capacity an abstraction over the actual commands required to perform a given task. If the project switched from using eslint to jshint, standard, or any other tool a contributor could continue using the same npm script and have the task run as expected.

Ultimately, though, I don't feel too strongly about this specific command. It seemed to me like it might be helpful, but if you disagree we can close out this PR.

Rob--W (approved):
No strong opinions either. It is simple and an improvement, so lgtm.

PR #580, opened 2025-04-04
Replace Markdown formatting with HTML

Josh-Cena:
This is showing up as raw markdown, so we should use plain HTML instead

### DIFF
--- a/examples.json
+++ b/examples.json
@@ -157,7 +157,7 @@
     "name": "discogs-search"
   },
   {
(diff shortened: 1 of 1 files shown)


Rob--W (approved):
It looks like the content is indeed interpreted as HTML, per https://github.com/mdn/rari/blob/fc9651804308ced6db52409d2bb1f8af4b2d1505/crates/rari-doc/src/templ/templs/webext_all_examples.rs#L25

Josh-Cena:
You would need to merge it btw, I don't have write access on this repo

PR #583, opened 2025-05-09
Prevent chrome API access

erosman:
`chrome.runtime.connect` & `chrome.runtime.sendMessage` are still accessible to userscripts.


Rob--W (commented):
What's the motivation for this change?

In Firefox, `chrome` is NOT defined in a user script context.

erosman:
> In Firefox, `chrome` is NOT defined in a user script context.

I saw it defined during testing.

image

Rob--W (approved):
Oh you're right. I forgot that I ended up adding it for compatibility.

PR #586, opened 2025-06-07
Add an example to search for multi-byte pattern in an array

def00111:


def00111:
@Rob--W Can you review this?

PR #592, opened 2025-07-10
Add an item to group the active tab in a new group.

steven-gardiner:
### Description

Modifies the tabs, tabs, tabs extension to demonstrate the usage of the tabs.group method.

### Motivation

The extension demonstrates other tabs-related methods but the group method is missing.

### Additional details

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/group

N/A


rebloor (approved):
Thanks @steven-gardiner

PR #617, opened 2026-01-20
Migrate native-messaging to MV3

rebloor:
### Description

This PR migrated the native-messaging example to use Manifest V3. Also include some improvements to the readme file.

### Motivation

Manifest V3 is now the recommended manifest version for web extensions.

Related content changes on MDN https://github.com/mdn/content/pull/42872.


willdurand (approved):
r+wc, makes sure to fix the typo in the commit title (I fixed the PR title)

native-messaging/add-on/background.js
Rob--W: In the mdn content article I already mentioned that the example is incompatible with Chrome due to the use of `background.scripts` without `background.service_worker`.

Besides that, this updated example demonstrates an anti-pattern.

In MV2 background scripts are long-lived, but in MV3 they are short lived, except when there is extension activity, including when a native messaging application is launched.

Instead of encouraging devs to have a long lived port + native messaging app that would keep their background script alive indefinitely, I recommend to put the logic in the action popup. Then you can render the response from the native app in the popup instead of console.log.

rebloor: @Rob--W update to use the popup, but not make Chrome compatible due to the extension ID issue.

PR #631, opened 2026-05-12
Replace `contextMenus` with `menus`

caugner:
### Description

Replace `contextMenus` with `menus`.

### Motivation

Resolves `webextallexamples` flaws reported by `@mdn/rari`.

### Additional details

See: https://caugner.github.io/mdn-flawless/#locale=en-US&templ=webextallexamples

Part of https://github.com/mdn/fred/issues/1462.

See also:

- https://github.com/mdn/content/pull/44041
- https://github.com/mdn/content/pull/44090

### DIFF
--- a/examples.json
+++ b/examples.json
@@ -106,8 +106,8 @@
   {
     "description": "Add a context menu option to links to copy the link to the clipboard, as plain text and as a link in rich HTML.",
     "javascript_apis": [
-      "contextMenus.create",
-      "contextMenus.onClicked",
+      "menus.create",
+      "menus.onClicked",
       "tabs.executeScript"
     ],
     "name": "context-menu-copy-link-with-types"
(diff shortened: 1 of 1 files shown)


Rob--W (changes requested):
These examples use contextMenus, not the menus APIs.

Restating what I wrote at https://github.com/mdn/content/pull/44090#pullrequestreview-4272308900 :

Although mdn only features "menus" without "contextMenus", we should really be documenting both instead of deleting references to contextMenus.

Sadly the same API is available under 2 namespaces, but different permissions.

contextMenus namespace + contextMenus permission

menus namespace + menus permission.

menus is available in Firefox and Safari, not Chrome.

contextMenus is available in all.

rebloor:
@rebloor (note to self) when this is resolved, if we are retaining `contextMenus`, confirm that these examples show on `menus.create` and `menus.onClicked` (which they don't at the moment) and, if necessary, add links.

PR #634, opened 2026-05-27
menu-demo to Manifest v3

rebloor:
### Description

This change migrates the menu-demo example to manifest V3. To do this, these changes were made:

**manifest.json**

-   `manifest_version`  upgraded from  `2`  to  `3`
-   Background script converted to non-persistent event page (`scripts`  array)
-   Added  `"scripting"`  permission (required by the Scripting API)
-   Extension ID added:  `menu-demo@mozilla.org`
-   `strict_min_version`  updated from  `56.0a1`  to  `109.0`  (minimum for MV3)
-   `data_collection_permissions`  key added with  `"required": ["none"]`
-   `open_at_install: false`  added to  `sidebar_action`  to suppress automatic sidebar opening on install

**background.js**

-   All menu item creation moved inside  `browser.runtime.onInstalled.addListener()`  — the correct pattern for non-persistent backgrounds, since menus persist across restarts but the top-level code would otherwise re-run each time the background wakes
-   `browser.tabs.executeScript`  replaced with  `browser.scripting.executeScript`  using  `func`  and  `args`  instead of an inline code string
-   `let blue/green`  (code strings) replaced with  `const blue/green`  (border value strings)
-   `removing.then(...)`  inlined to  `browser.menus.remove(...).then(...)`
-   Minor typo fix in a comment: "only work" → "only works"

**examples.json**

-   `runtime.lastError`  replaced with  `runtime.onInstalled`
- `tabs.executeScript` replaced with `scripting.executeScript`

**messages.json**

- Changed the sidebar menu label to "Toggle sidebar" to more accurately reflect behavior

**README.md**
- Clarified that additional changes, beyond the use of the `contextMenus` alias are needed to run on Chrome.
- minor edits.

### DIFF
--- a/examples.json
+++ b/examples.json
@@ -338,6 +338,7 @@
       "menus.onClicked",
       "menus.remove",
       "menus.update",
+      "runtime.lastError",
       "runtime.onInstalled",
       "scripting.executeScript"
     ],
--- a/examples.json
+++ b/examples.json
@@ -338,8 +338,8 @@
       "menus.onClicked",
       "menus.remove",
       "menus.update",
-      "runtime.lastError",
-      "tabs.executeScript"
+      "runtime.onInstalled",
+      "scripting.executeScript"
     ],
     "name": "menu-demo"
   },
--- a/menu-demo/background.js
+++ b/menu-demo/background.js
@@ -1,16 +1,4 @@
 /*
-Called when the item has been created, or when creation failed due to an error.
-We'll just log success/failure here.
-*/
-function onCreated() {
-  if (browser.runtime.lastError) {
-    console.log(`Error: ${browser.runtime.lastError}`);
-  } else {
-    console.log("Item created successfully");
-  }
-}
-
-/*
 Called when the item has been removed.
 We'll just log success here.
 */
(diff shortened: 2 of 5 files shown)


menu-demo/background.js
Rob--W: When `command` is assigned, `onClicked` only fires if not supported.
```suggestion
      console.warn("_execute_sidebar_action not supported");
```

menu-demo/background.js
```diff
@@ -1,15 +1,3 @@
-/*
-Called when the item has been created, or when creation failed due to an error.
-We'll just log success/failure here.
-*/
-function onCreated() {
-  if (browser.runtime.lastError) {
-    console.log(`Error: ${browser.runtime.lastError}`);
```
Rob--W, line 7: `menus.create` is special, unlike many other extension APIs it does not return a `Promise` even though it is async.

I think that it would still be useful to log an error to show how error handling works. The `"Item created successfully"` part can be dropped though.

The `console.log` call should include a `.message` though, as in: `browser.runtime.lastError.message`

examples.json
```diff
@@ -338,8 +338,8 @@
       "menus.onClicked",
       "menus.remove",
       "menus.update",
-      "runtime.lastError",
```
Rob--W, line 341: Keep `runtime.lastError` as an example, see my comment in background.js.

Rob--W (approved):
Looks good with minor adjustments.

menu-demo/background.js
rebloor: @Rob--W done, plus some tweaks to the comments

LeoMcA (approved):
examples.json change looks fine from from engineering perspective

Rob--W (approved):

PR #651, opened 2026-09-08
ci(lint): pin Node.js via `.nvmrc` instead of `latest`

caugner:
### Description

Add a `.nvmrc` pinning Node.js 24, and use `node-version-file` instead of `node-version: "latest"` in both `lint.yml` jobs.

### Motivation

Prevent the lint jobs from silently moving to a new Node.js major, and with it a new bundled npm. npm 12 changes several `npm install` defaults, and npm has asked for it to be backported into the active LTS lines, so `latest` would pick it up without any change on our side.

### Additional details

This brings the repo in line with the rest of the MDN repos, which all use `node-version-file: .nvmrc`.

`npm ci` here has no dependencies with install scripts, so no `allowScripts` allowlist is needed. Verified against `npm@12.0.2` with `npm ci --strict-allow-scripts`.

The example projects in subdirectories have no lockfiles and are not installed by CI, so they are untouched.

Part of mdn/fred#1868.


LeoMcA (approved):

caugner:
@mdn/add-ons As default owner, changes to files in the root folder (like this new `.nvmrc` file) need approval from you, though we could add an exception for this file, given that `@mdn/engineering` bumps Node versions across MDN repositories.

Rob--W (approved):

